PR #133, opened 2023-05-16
Added typescript

TheRizz:
Ran the npm install command with node version 14.17.0 and npm version 6.14.13. Installed dependencies for using typescript in this project. Updated the readme to include an explanation on node versions and a download link. Updated frontend app files from js to tsx files.
Tsconfig file added as a result of initializing typescript in the repo.


misslame (approved):
LGTM!

README.md
```diff
@@ -1,3 +1,12 @@
+# Before Getting Started
+
+Please be sure when running the front end repository that you have the correct versions of npm and node.
+The correct version can be found and installed [here](https://nodejs.org/download/release/v14.17.0/).
```
rezatnoMsirhC, line 4: Node 14 is no longer supported as of April 30. We should upgrade.

tsconfig.json
```diff
@@ -0,0 +1,110 @@
+{
```
rezatnoMsirhC, line 1: 1. Might be a good idea to extend from node tsconfig (i.e. [@tsconfig/node20](https://www.npmjs.com/package/@tsconfig/node20))
2. I think we should specify `outDir` and avoid checking-in the `.js` and `.d.ts` generated from typescript-compile
3. How are you commenting out anything in this file? JSON doesn't support comments afaik. Is tsconfig technically not JSON-compliant? Also commenting so many lines in one file is messy imo, but it might help the newbies discover more options.

README.md
```diff
@@ -1,3 +1,12 @@
...
+
+To check if your versions are correct you can use the following commands:
+Check node version with `node -v` -> This should show v14.17.0 in the console
+Check npm version with `npm -v` -> This should show 6.14.x in the console
```
rezatnoMsirhC, line 8: Why not the latest, stable version of npm?

misslame:
Chris better at dev ops related reviews than me it seems

README.md
TheRizz: I will test out the latest release and update the readme after confirming that everything is building and running well.

tsconfig.json
TheRizz: 1. The tsconfig was generated from the older version of node. I will generate a new version of the tsconfig after the version of node has been installed. (seems like the tsconfig generated is based on the version of node, never knew that, but cool to know)

2. Typically I add the generated output directory to the .gitignore for the repo so we dont push any big, bulky compiled files into the repo when they are not needed. Typically building the production version is part of the deployment process. Not sure how we plan to handle this in our pipeline.

3. Also part of the automatically generated tsconfig file. I can remove the comments if there are any present in the new version, but it is also nice to be able to see the different settings available for those who were not involved in the dev ops aspect of this project if they wanted to learn.

README.md
TheRizz: Will swap to latest release of node. Looks like this is version 20.2.0 right now.

README.md
TheRizz: Updated readme.

tsconfig.json
TheRizz: Updated tsconfig.
To add files to git ignore in future.

README.md
TheRizz: Updated

rezatnoMsirhC (approved):
LGTM

Issue #127, opened 2023-05-16
G1.3 Init Profiles Table


rezatnoMsirhC:
Write a SQL query to define the schema of and create the Profiles table.

This is the current schema:
- accountID (primary/foreign key)
- name
- preferredName
- preferredPronouns
- biography
- profilePicture

Refer to [the UML-Database diagrams.net Page](https://app.diagrams.net/#G18i2FGzLBHPOk-yKIdq-dVpT0CnBap6Ri) for the latest details.

Issue #125, opened 2023-05-16
G1.1 Init Accounts Table


rezatnoMsirhC:
Write a SQL query to define the schema of and create the Accounts table.

This is the current schema:
- accountID (primary key)
- email
- encryptedPass
- phoneNum
- displayTheme
- connectedAccounts
- enable2Factor
- isMentor
- isMentee

Refer to [the UML-Database diagrams.net Page](https://app.diagrams.net/#G18i2FGzLBHPOk-yKIdq-dVpT0CnBap6Ri) for the latest details.

misslame:
peppa

Issue #124, opened 2023-05-16
G1 Database


rezatnoMsirhC:
## Details:
We are persisting app and user data in a SQL database

### Acceptance Criteria:
- [x] ¯\\\_(ツ)\_/¯
- [x] Create a database with the below tables

### Technical Tasks:
- [x] #125
- [x] #126
- [x] #127
- [x] #128
- [x] #129
- [x] #130
- [x] #131
- [x] #132

PR #143, opened 2023-05-28
Leaf/1 3 init profiles table

peppacaiou:
Note, the database isn't set up, only the the accountable and the profile SQL files are the ones that should be reviewed for this pull request
#125 
#127 
#124


.gitignore
```diff
@@ -25,4 +25,7 @@ yarn-error.log*
 # django stuff
 *.env
 *.pyc
-*.sqlite3
\ No newline at end of file
+*.sqlite3
+
+#mysql stuff
```
misslame, line 30: How is vscode "#mysql stuff"? Could you provide either an appropriate comment or no comment at all.

backend/accounts.sql
```diff
@@ -0,0 +1,13 @@
+-- @block 
+CREATE TABLE IF NOT EXISTS account (
```
misslame, line 2: How is this run? Is it ran on start up?

misslame (changes requested):
I have thoughts

backend/settings.py
```diff
@@ -85,8 +85,17 @@
...
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': BASE_DIR / 'db.sqlite3',
+        'NAME': 'user_data',
+        'ENGINE': 'django.db.backends.mysql',
+        'HOST': '127.0.0.1',
+        'PORT': 3306,
+        'USER': 'root',
+        'PASSWORD': 'lmaozedong',  # change this to env variable maybe
+        'OPTIONS': {
+          'autocommit': True,
```
rezatnoMsirhC, line 95: I'm not sure if we'll be using any atomic-transactions, but it might be good to note that this should be switched to `False` if so. Consider reading [this MySQL documentation about autocommit, Commit, and Rollback](https://dev.mysql.com/doc/refman/8.0/en/innodb-autocommit-commit-rollback.html)

backend/settings.py
```diff
@@ -85,8 +85,17 @@
...
+          'use_oure': True,
```
rezatnoMsirhC, line 96: Should this be `use_pure` instead?

backend/settings.py
```diff
@@ -85,8 +85,17 @@
...
+          'init_command': "SET foo='bar';"
```
rezatnoMsirhC, line 97: Is there a follow-up task which captures setting a real `init_command`? I think this should also have a `TODO`.

backend/profileTable.sql
```diff
@@ -0,0 +1,14 @@
+-- @block
+CREATE TABLE IF NOT EXISTS profileTable(
+    accountID INT NOT NULL,
```
rezatnoMsirhC, line 3: Would it be possible to specify `PRIMARY KEY` on this line instead of separately on line 11? That would eliminate the need for `NOT NULL` here.

backend/accounts.sql
```diff
@@ -0,0 +1,13 @@
...
+    accountID INT PRIMARY KEY NOT NULL AUTO_INCREMENT,
```
rezatnoMsirhC, line 3: `NOT NULL` is redundant because `PRIMARY KEY` cannot be null.

backend/accounts.sql
```diff
@@ -0,0 +1,13 @@
...
+    email VARCHAR(255) NOT NULL,
+    encryptedPass VARCHAR(255) NOT NULL,
```
rezatnoMsirhC, line 5: If we're using `SHA256` encryption, we would only need `numCharacters = 256 (bits) / numBitsInRepresentation`. For example, if we're going to represent the encrypted-password in hexadecimal (4 bits per digit), we'd only need `256 / 4 = 64` characters.

backend/profileTable.sql
```diff
@@ -0,0 +1,14 @@
...
+    userName VARCHAR(255) NOT NULL,
```
rezatnoMsirhC, line 4: Is this the user's real name (i.e. `Bob Hasalastname`) or their display name (i.e. `pussyslayer69`)? If it's a display name, we might want to consider adding a column for their real name either here or in the accounts table.

backend/profileTable.sql
```diff
@@ -0,0 +1,14 @@
+-- @block
+CREATE TABLE IF NOT EXISTS profileTable(
```
rezatnoMsirhC, line 2: `profileTable` -> `Profiles`

backend/accounts.sql
rezatnoMsirhC, line 2: `account` -> `Accounts`

backend/databse.py
rezatnoMsirhC, line 1: `databse.py` -> `database.py`

backend/profileTable.sql
rezatnoMsirhC, line 1: `profileTable.sql` -> `profiles.sql`

backend/accounts.sql
```diff
@@ -0,0 +1,13 @@
+-- @block 
+CREATE TABLE IF NOT EXISTS account (
+    accountID INT PRIMARY KEY NOT NULL AUTO_INCREMENT,
+    email VARCHAR(255) NOT NULL,
```
rezatnoMsirhC, line 4: Based on [Errata 1690](https://www.rfc-editor.org/errata/eid1690), email addresses are limited to 254 characters.

misslame:
Where are the linked tasks?

backend/accounts.sql
peppacaiou: Thsi is simply the mysql command, the actual database comes later

backend/accounts.sql
```diff
@@ -0,0 +1,13 @@
...
+    encryptedPass VARCHAR(255) NOT NULL,
+    displayTheme int NOT NULl,
+    phoneNum VARCHAR(20),
```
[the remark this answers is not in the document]
peppacaiou: It seems that the longest is 15 including country code, so I will probably keep it at 15. I kept it null in case we wanted to allow email as the sole identifier.

backend/profileTable.sql
peppacaiou: From what I've tried it doesn't work. Removing NOT NULL works fine if that's good.

backend/accounts.sql
peppacaiou: I'm not really familiar with how encryptions work but what if we decide to us e different encryption? Or is that extra space not necessary for any kind of encryption.

backend/profileTable.sql
peppacaiou: name is a reserved keyword that I can't use so I just chose username. Is it ambiguous? Should I use actualName or something like that.

backend/profileTable.sql
```diff
@@ -0,0 +1,14 @@
...
+    accountID INT NOT NULL,
+    userName VARCHAR(255) NOT NULL,
+    preferredName VARCHAR(255),
+    preferredPronun VARCHAR(255),
+    biography TEXT(500),
```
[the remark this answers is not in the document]
peppacaiou: Slay

backend/accounts.sql
rezatnoMsirhC: Different encryption algorithms will produce varying output lengths. We can leave this as-is for now and revisit once we know what's required. If the encryption algorithm we use produces an output longer than 255 characters, things will break here.

backend/profileTable.sql
rezatnoMsirhC: Huh that's weird. Admittedly I'm not very familiar with SQL schemata, so ideally I'd want someone else more familiar to chime-in. I suppose this is fine though.

backend/profileTable.sql
rezatnoMsirhC: I think you can leave this as-is. I think it would be nice to have a sample record as a comment in these files, or maybe in some other reference document we're tracking for this project.

Issue #142, opened 2023-05-23
B4 Login Form


Darren52BT:
As a user, I should be able to login with my email and password, sign in with google or discord, or be directed to the page to retrieve my password if I have forgotten it. 

Details: Follow the figma design of the login form component within the page designs such as the logged out home screen and the login page, and the constraints listed by the login page issue.

Technical Details:

+ React
+ UI / UX
+ Front-end

Related Issues:
#114 
#135

Issue #114, opened 2023-04-25
B4.1	Login page


GuyWhoCode:
Story: **As a User (Mentor, Mentee, or Educator), I want to be able to log in with my account so I can start using the service.**

Details: Create a login page with text inputs for email and password as well as a button to submit. Include buttons for 3rd party logins like Google or Facebook. Follow Figma. Ensure all forms are correlated to the correct inputs and passwords are hidden on the screen. 

Technical Details:
- React
- UI /  UX
- Front-end

PR #166, opened 2023-05-30
Leaf/b4.1 login page

Leaversa:
addresses #114 and #142


src/components/LoginForm.tsx
```diff
@@ -0,0 +1,42 @@
...
+              OR
+            </Typography>
+          </Stack>
+        </Grid>
+      </Grid>
+      <Grid item xs>
+        <Stack spacing={2} sx={{ alignItems: 'center' }}>
+          <Button sx={{ bgcolor: '#D9D9D9', color: 'black', '&:hover': { bgcolor: '#D9D9D9' } }}>SIGN IN WITH GOOGLE</Button>
+          <Button sx={{ bgcolor: '#748ADA', color: 'white', '&:hover': { bgcolor: '#748ADA' } }}>SIGN IN WITH DISCORD</Button>
+        </Stack>
```
PattyCakesxD, line 36: The size of these buttons are too small at 100% size on my end, so the text gets wrapped and it looks very weird. They also don't have the Google and Discord logos, you should add them while you're at it.

image

src/components/LoginForm.tsx
```diff
@@ -0,0 +1,42 @@
...
+        <Grid item xs>
+          <Typography variant="h2" fontWeight={'bold'} sx={{ marginLeft: 'auto' }}>
+            Login
+          </Typography>
+        </Grid>
+        <Grid item xs sx={{ width: '70%' }}>
+          <Stack spacing={2}>
+            <TextField type="text" label="Email" />
+            <TextField type="password" label="Password" />
+          </Stack>
```
PattyCakesxD, line 16: Following the figma, the Login text should be left-aligned, and the textfields should be the same length as the Login button. I'm assuming you coded this in with a smaller screen size since minimizing the screen makes it look fairly identical to the figma, but you need to account for different screen sizes for UX purpose

image

src/components/LoginForm.tsx
```diff
@@ -0,0 +1,42 @@
...
+        </Grid>
+        <Grid item xs>
+          <Stack spacing={2}>
+            <Button variant="contained" sx={{ width: '15em' }}>
```
PattyCakesxD, line 20: These buttons should be the lime-ish color like in the figma. I saw your messages on discord about the theming though, so not sure if it's necessary.

It'd be cool if you hardcoded it in for now though, would look nicer than it currently does and it would be a good reference to what the final page would look like in the end.

src/components/LoginForm.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
```
PattyCakesxD, line 1: The cocogoose font is actually in the main branch and not in the dev for some reason, so if you want to change the font to the figma then the file is in public/fonts/cocogoose.tff

PattyCakesxD (changes requested):
Pretty much all the comments are tied to differences in the figma and some misc styling, nice progress so far!

src/components/LoginForm.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { Button, Grid, Link, Stack, Typography, TextField } from '@mui/material';
+const LoginInput = () => {
+  return (
```
owwix, line 4: set up state hooks to store form data

owwix (changes requested):
set up state hooks to store form data, pattycakesxd already addressed the styling

src/components/LoginForm.tsx
Leaversa: @TheRizz told me I should use the default mui color scheme for now until theming gets properly set up.

src/components/LoginForm.tsx
Leaversa: @TheRizz mentioned that icons would be added later through an external library and that I could omit them for now.

src/components/LoginForm.tsx
Leaversa: I'll choose not to include the font for the current PR since I assume it'll be implemented into the typography once theming is added.

owwix (approved):
good changes

PattyCakesxD (approved):
Consider implementing the cocogoose font before merging, but it looks great otherwise

PR #173, opened 2023-06-01
Configure axios with django

owwix:
- configured axios on the frontend
- configured django server settings and dependencies to accept axios requests
- temporarily disabling csrf protection middleware in django server


misslame:
@owwix  retarget this to dev branch not main please

PR #176, opened 2023-06-02
Leaf/1.6 tags bridge tables

sosropartono:
Created SQL queries for tags and tags bridge tables

#125


backend/settings.py
```diff
@@ -85,8 +85,17 @@
...
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': BASE_DIR / 'db.sqlite3',
+        'ENGINE': 'django.db.backends.mysql',
+        'NAME': 'user-data',
+        'HOST': '127.0.0.1',
+        'PORT': 3306,
+        'USER': 'root',
+        'PASSWORD': 'lmaozedong',
```
PattyCakesxD, line 93: Is this password meant to be here? (similar to the other table pr)

PattyCakesxD (approved):
Just 1 concern, lgtm otherwise

mesh/profiles/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path, include
+from . import views as profile_views
+from django.http import HttpRequest
+
+urlpatterns = [
+    
+    # Home Path for Example API 'exampleapi'
```
misslame, line 7: This is not the example api urls

mesh/profiles/urls.py
```diff
@@ -0,0 +1,10 @@
...
+    path('biography/', profile_views.biography, name="biography"),
```
misslame, line 8: What is this biography stuff? How is it related to tags table?

misslame (changes requested):

mesh/profiles/urls.py
sosropartono: Sorry about that, I do not remember replacing it with that... will fix

PR #178, opened 2023-06-02
Leaf/1.7 notifications table

sosropartono:
Notification Tables added, still cannot test the database, only provided SQL files


backend/settings.py
```diff
@@ -85,8 +85,17 @@
...
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': BASE_DIR / 'db.sqlite3',
+        'ENGINE': 'django.db.backends.mysql',
+        'NAME': 'user-data',
+        'HOST': '127.0.0.1',
+        'PORT': 3306,
+        'USER': 'root',
+        'PASSWORD': 'lmaozedong',
```
PattyCakesxD, line 93: Is this password meant to be here?

PattyCakesxD (approved):
Just 1 concern to be checked, lgtm otherwise

backend/settings.py
sosropartono: Yeah, this doesn't seem like the best practice so going to remove this portion just so we have the SQL files.

misslame (changes requested):
Update to use refactor code and make the model for the table using django

mesh/settings.py
```diff
@@ -140,9 +140,3 @@
 
 # Default primary key field type
 # https://docs.djangoproject.com/en/4.2/ref/settings/#default-auto-field
-
-DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'
-
-STATICFILES_DIRS = [
```
misslame, line 146: Why is this being deleted

misslame (changes requested):

mesh/settings.py
sosropartono: Sorry about that, was debugging and accidentally deleted these lines.

mesh/notifications/models.py
```diff
@@ -1,7 +1,11 @@
...
 # Create your models here.
 
 class Notification(models.Model):
-    #TODO
-    pass
\ No newline at end of file
+    accountID = models.ForeignKey(Account, on_delete=models.CASCADE)
+    notificationID = models.AutoField(primary_key=True)
+    timeStamp = models.TimeField()
+    notificationContextJSON = models.JSONField()
```
peppacaiou, line 10: what is notficationContext supposed to be about?

sosropartono: I followed what was on the table but I think its just information about the notification like the message/type of notification?

PR #179, opened 2023-06-02
Leaf/b3.2 prompt opt in login

PattyCakesxD:
Implements two main components:
- A popup (TwoFactorAuthModal) that prompts users to enable 2FA (should only be a first-time thing; new users)
- A banner (TwoFactorBanner) that reminds users about enabling 2FA (only if 2FA hasn't been enabled yet)

Note that the "2FA-Enabled" flag is temporary to simulate whether it's been enabled by the user or not. Right now, there isn't any code that plays around with it, so it'll always be a false value. If you want to test, you can manually give yourself that flag or just modify the code in 2FA-Remind.tsx.

Also note that this branch is using a temporary css to give the body a background color of the site's default green-ish color. You can delete this all you want, it's just used as a reference; the components themselves can be placed anywhere.

When the 2FA stuff is done, it should be very easy to connect everything together.


Shadowowl888 (approved):
2FA popup and banner looks good.

misslame:
This looks ready to merge! Hopefully, you can demo it :D @PattyCakesxD

misslame:
Remember, Squash and merge, and modify the commit message to include the issue number, for example: 

#[issue number]: B3.2 Implement Prompt Opt-in of two factor authentication banner UI.

Issue #109, opened 2023-04-25
B2.2	Forgot Password Form


GuyWhoCode:
Story: **As a User (Mentor, Mentee, or Educator), I want to be able to request a password change if I forgot my password.**

Details: Allows the user to type their new password. Requires the creation of 2 input boxes: one for the initial password and another for confirmation. When the form is submitted, the form should send a POST request with the email (or token) and the new password.

Technical Details:
- React
- HTTP REST Post
- UX

PR #181, opened 2023-06-02
Leaf/b2.2 forgot password form

owwix:
B2.2 Forgot Password Form #109


misslame:
There are a lot of commits here that don't seem related to your fix, where did you branch off of? you may need to do a rebase or cherry-pick. squashing/merging all of those commits together are bound to cause issues. We can't squash and merge things that aren't related to each other.

PR #183, opened 2023-06-02
Leaf/b2.2 forgot password form v2

owwix:
B2.2 Forgot Password Form https://github.com/LetsMesh/Site/issues/109

cherry picked some commits and opened new PR


src/fonts/fonts.css
misslame, line 1: I believe that we already have the font in Dev, could you double check and make sure you're not making a redundant change.

src/index.tsx
misslame, line 1: I think we need to not make changes to this file for now due to merge conflicts. Everyone seems to be making the same change to this file. We can tie your work into the index file eventually with Ryan M's task.

misslame (changes requested):
Small changes.

misslame (changes requested):
Small changes.

Issue #136, opened 2023-05-20
F4.1 Custom Footer appear on all pages


misslame:
Story: **As a product owner, I want to be able to ensure a welcoming environment on the site/app by providing information in a home screen footer**

Details: Provide the footer. Follow the figma design to the best of your ability. Fill any gaps with separate task issues following task format. 

Technical Details: 
- React
- Front-end
- MUI
- UI/UX

PR #189, opened 2023-06-06
Leaf/f4.1 custom footer appear on all pages

Darren52BT:
addresses #136.
I've added the custom footer component to the components folder, which is mainly for containing company information and related links. The links should be on the right side of the page listed in columns, though I set up the row of link lists to take up most of the footer width when the screen width goes below 600px. The links themselves currently have empty references; we need to add them later when we have the proper links set up.  The footer component is meant to be placed below page components so that it will show up at the bottom. Upon running the app, you should be able to see the login page and the footer at the bottom.

Edit: changed footer links to be slighter lighter than the header to increase contrast
image


src/Footer.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+import React from "react";
+// import {
+//   Box,
+//   Container,
+//   Row,
+//   Column,
+//   FooterLink,
+//   Heading,
+// } from "./FooterStyles"
```
rezatnoMsirhC, line 11: Please remove this

src/Footer.tsx
```diff
@@ -0,0 +1,134 @@
+// need to convert to typescript
```
rezatnoMsirhC, line 1: Please convert this to a `TODO` and add a corresponding issue link.
(i.e. `// TODO: need to convert to typescript (https://github.com/LetsMesh/Site/issues/42069)`)

src/Footer.tsx
```diff
@@ -0,0 +1,134 @@
...
+      h3: {
+        fontSize: "13px",
+        fontWeight: "400",
+      },
+    },
+  });
+
+  return (
+    <ThemeProvider theme={theme}>
+      {/*footer*/}
```
rezatnoMsirhC, line 60: Is an empty `{ }` required here? idk much about react

src/Footer.tsx
```diff
@@ -0,0 +1,134 @@
...
+              <Link href="">Discord Inv</Link>
+              <Link href="">Report Bug</Link>
+            </Stack>
+            <Stack direction="column" rowGap={1}>
+              <Typography variant="h2">Related Links</Typography>
+              <Link href="">Myspace</Link>
+              <Link href="">Hotline</Link>
+              <Link href="">More Ram</Link>
+              <Link href="">Leetcode</Link>
+            </Stack>
```
rezatnoMsirhC, line 107: We should have a `TODO` here to add the actual links. Please create an issue for this if there isn't already one.

src/index.tsx
```diff
@@ -1,16 +1,18 @@
...
-
+import Login from "./pages/Login";
+import Footer from "./Footer";
 const root = ReactDOM.createRoot(document.getElementById("root")!);
 root.render(
   <React.StrictMode>
-    <TwoFactorAuthReminders />
+    {/* <TwoFactorAuthReminders /> */}
+    <Login />
+    <Footer />
```
rezatnoMsirhC, line 15: Why was `<TwoFactorAuthReminders />` removed? If we really don't need it, then I think the import above for it should be removed.

rezatnoMsirhC:
Do you have any screenshots to show what exactly changed here? Ideally there would be a before and after for the reviewers to look at.

src/Footer.tsx
Darren52BT: Oh my bad those were old comments that i forgot to delete

src/Footer.tsx
Darren52BT: i was just adding some comments to document the sections

Darren52BT:
image

When you run the app you should be able to see the new footer at the bottom of the login page.

src/components/Footer.tsx
```diff
@@ -0,0 +1,126 @@
...
+      h3: {
+        fontSize: "13px",
+        fontWeight: "400",
+      },
+    },
+  });
+
+  /**
+   * @TODO : add actual links to the link elements (https://github.com/LetsMesh/Site/issues/191)
+   * */
```
rezatnoMsirhC, line 50: Ideally this `TODO` would go directly above the associated code. If we switch to HTML-style comments, we can move this to just above line 79.

src/Footer.tsx
Darren52BT: According to this, html comments are considered as DOM nodes in JSX so we need to wrap comments in curly braces. 
https://wesbos.com/react-jsx-comments

src/Footer.tsx
rezatnoMsirhC: So we don't want this written in TypeScript? I prefer it over JavaScript, but I'll let Ryan chime in on what's better. Ideally we should have all JavaScript or all TypeScript imo.

rezatnoMsirhC (changes requested):

src/Footer.tsx
Darren52BT: So that was a comment left by the person who had this task previously where he did it in JS, I overwrote the file into typescript but I forgot to remove his comment, should be gone now

rezatnoMsirhC: Ah okay gotcha. I'm not used to React at all.

src/Footer.tsx
rezatnoMsirhC: 😢

meshapp/src/components/Footer.tsx
```diff
@@ -0,0 +1,126 @@
...
+          <Grid
+            container
+            direction="row"
+            justifyContent="space-around"
+            xs={11}
+            sm={6}
+            sx={{ flexWrap: "nowrap" }}
+          >
+            {/**
+             * @TODO : add actual links to the link elements (https://github.com/LetsMesh/Site/issues/191)
```
misslame, line 77: Create another issue / task for this so it isn't lost in code :D (put the reference of the TODO in the issue description)

misslame (approved):
Looks like we can add it to any page I am assuming. 

Thanks for updating index to show more.

meshapp/src/components/Footer.tsx
Darren52BT: Isn't this the issue I linked in the TODO? what does the reference of the TODO exactly mean?

owwix (commented):
make headings a slightly different color than the footer links to increase contrast and readability

calhuy (approved):
looks like it works as intended, I also made the screen smaller and it seemed to be fine. Cool!

rezatnoMsirhC (approved):

Issue #126, opened 2023-05-16
G1.2 Init Settings Table


rezatnoMsirhC:
Write a SQL query to define the schema of and create the Settings table.

This is the current schema:
- poopy (Poopmary key)

Refer to [the UML-Database diagrams.net Page](https://app.diagrams.net/#G18i2FGzLBHPOk-yKIdq-dVpT0CnBap6Ri) for the latest details.

misslame:
This requires some Design work

JosephKahale:
@rezatnoMsirhC Currently working on adding option for unallowed words / phrases into settings table, just wondering if we are allowing user to add unallowed words individually, or if its just some toggle in the settings that enables some filter?

rezatnoMsirhC:
@JosephKahale I think allowing users to submit their own words/phrases would get too complicated, so let's just define them internally for now.

JosephKahale:
@rezatnoMsirhC Sounds good, is there specific reasoning on displayTheme and enable2Factor being in Accounts table and not in a settings table?

rezatnoMsirhC:
@JosephKahale Probably because we wrote the Accounts table first, and thought about the Settings table most recently without defining a schema. Those should probably be in the Settings table. There are probably some other columns which aren't placed quite correctly as we've listed them, so feel free to move them around and provide your justification.

JosephKahale:
Sounds good! Thanks for the clarification! @rezatnoMsirhC

JosephKahale:
#197

PR #197, opened 2023-06-10
G1.2 Init Settings Table

JosephKahale:
#126 
Created table for accountSettings and blockedAccountBridge.

AccountSettings: General account settings
BlockedAccountBridge: Creates a composite pk and fk's to link blocker's blocked accounts to blocker's account


mesh/accountsettings/accountsettings.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS AccountSettings(
+	settingID INT PRIMARY KEY AUTO_INCREMENT,
+    accountID INT,
+    isVerified BOOL NOT NULL DEFAULT FALSE,
+    enableContentFilter BOOL NOT NULL DEFAULT FALSE,
+    displayTheme ENUM("light","dark") NOT NULL DEFAULT "light",
+    is2FAEnabled BOOL NOT NULL DEFAULT FALSE,
+    phoneNum VARCHAR(15),
```
rezatnoMsirhC, line 8: Have you checked with @peppacaiou about taking these away from the `Accounts` table? Their PR #143 already captures these in `Accounts`.

mesh/accountsettings/accountsettings.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS AccountSettings(
+	settingID INT PRIMARY KEY AUTO_INCREMENT,
```
rezatnoMsirhC, line 2: 1. Do we need a unique ID for settings other than the `FOREIGN KEY accountID`? I believe this would only be useful if we wanted to be sharing a set of settings amongst several accounts/users, which I don't believe to be the intention here. 
1. nitpick: there's an extra indentation here

mesh/accountsettings/blockedaccountbridge.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS BlockedAccountBridge(
+	settingID INT,
```
rezatnoMsirhC, line 2: 1. I think this can be `blockerAccountID` if we decide to remove `settingID` from `AccountSettings` (*refer to my other comment about the necessity of `settingID`*). Examples of this relationship:
    1. Bob (1) has blocked Chris (2) 
        1. `blockerAccountID = 1`
        2. `blockedAccountID = 2`
    1. Bob (1) has blocked Tony (3)
        1. `blockerAccountID = 1`
        2. `blockedAccountID = 3`  
1. nitpick: there's an extra indentation here

mesh/accountsettings/accountsettings.sql
```diff
@@ -0,0 +1,11 @@
...
+    accountID INT,
+    isVerified BOOL NOT NULL DEFAULT FALSE,
```
rezatnoMsirhC, line 4: PR #143 uses `BOOLEAN` and `DEFAULT 0`, while you use `BOOL` and `DEFAULT FALSE`. They are semantically-identical, but I think we should be consistent for clarity. `FALSE` is more readable than `0` imo, but `BOOLEAN` vs. `BOOL` probably doesn't matter.

mesh/accountsettings/accountsettings.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS AccountSettings(
```
rezatnoMsirhC, line 1: nitpick: tbh I think this should be called `Settings` as I don't think prefacing with `Account` adds any clarity

rezatnoMsirhC (changes requested):

mesh/accountsettings/accountsettings.sql
```diff
@@ -0,0 +1,11 @@
...
+	settingID INT PRIMARY KEY AUTO_INCREMENT,
+    accountID INT,
+    isVerified BOOL NOT NULL DEFAULT FALSE,
+    enableContentFilter BOOL NOT NULL DEFAULT FALSE,
```
rezatnoMsirhC, line 5: For clarity, can you please rename this to something like `hasContentFilterEnabled`? I think the `is`, `has`, `does`, etc. prefixes convey the `BOOL` typing of this column best.

mesh/accountsettings/accountsettings.sql
JosephKahale: Just checked, good to go!

mesh/accountsettings/accountsettings.sql
JosephKahale: @rezatnoMsirhC I do agree with you here, so we can identify settings just plainly on foreign key in the table.

JRSiwiecki:
Just to clarify, the migrations folder & files are automatically generated by Django right?

JosephKahale:
> Just to clarify, the migrations folder & files are automatically generated by Django right?
I wouldnt say automatically generated because there are a few commands I have to run to get them migrated. But I did some research and from my understanding its best to commit the migrations. If im mistaken please correct me :)

JRSiwiecki (approved):
I see, then otherwise this looks good to me!

PR #206, opened 2023-06-16
Leaf/a2.5 http post call

### LINKED ISSUE #12: A2.5 HTTP Post Call Handling
Story: ﻿As a User (Mentor, Mentee, or Educator), I want to be able to provide a biography description to describe myself, my goals, & my interests.﻿
﻿
﻿﻿Details:
﻿Allow for the retrieval of a new biography to be stored and for the previous biography to be overwritten.

﻿Technical Information:
- Django
- Handle Post Call
- Error Handling as a response is expected for improper Post Calls
- Back-end

Leaversa:
resolves #12 

Does not actually store biography into the database. That will be solved with #9.


mesh/accounts/views.py
```diff
@@ -1,3 +1,16 @@
...
+from django.http import JsonResponse
+from rest_framework.views import exception_handler
 
-# Create your views here.
+@csrf_exempt
+def bio_view(request):
+    if request.method == "POST":
+        data = request.POST
+        if 'biography' not in data:
+              return JsonResponse({'error': 'Invalid request. Missing biography field.'}, status=400)
```
peppacaiou, line 10: maybe use status code 412

mesh/accounts/views.py
```diff
@@ -1,3 +1,16 @@
...
+        else:
+            print(data.get('biography'))
```
peppacaiou, line 12: I don't see a logic to save the data.

mesh/accounts/views.py
```diff
@@ -1,3 +1,16 @@
...
+            # Save Biography to Database
+            return JsonResponse({'message': 'biography saved successfully'}, status=200)
```
peppacaiou, line 14: wrong status code, should be 201

mesh/accounts/views.py
```diff
@@ -1,3 +1,16 @@
...
+    else:
+        return JsonResponse({'error': 'Method not allowed'}, status=405)
```
peppacaiou, line 16: better error message may be to specify which method they used rather than just "method not allowed". e.x. get method not allowed.

peppacaiou (commented):
I have thoughts

mesh/accounts/views.py
```diff
@@ -1,3 +1,16 @@
-from django.shortcuts import render
+from django.views.decorators.csrf import csrf_exempt
+from django.http import JsonResponse
+from rest_framework.views import exception_handler
 
-# Create your views here.
+@csrf_exempt
+def bio_view(request):
+    if request.method == "POST":
```
peppacaiou, line 7: you need some sort of way to determine which biography to change

peppacaiou (approved):
Lgtm (luigi got them moves)

mesh/profiles/views.py
```diff
@@ -1,3 +1,16 @@
...
+@csrf_exempt
+def bio_view(request):
+    if request.method == "POST":
+        data = request.POST
+        if 'accountID' not in data:
+            return JsonResponse({'error': 'Invalid request. Missing accountID field.'}, status=401)
+        if 'biography' not in data:
+            return JsonResponse({'error': 'Invalid request. Missing biography field.'}, status=400)
+        else:
+            print(data.get('accountID') + " " + data.get('biography')) # TODO: Save to database
```
misslame, line 13: Make issue for the TODO link in PR description that this introduces that issue

misslame (approved):
Some thoughts but will approve. Trusting that you'll take comments seriously.

mesh/profiles/views.py
```diff
@@ -1,3 +1,16 @@
-from django.shortcuts import render
+from django.views.decorators.csrf import csrf_exempt
+from django.http import JsonResponse
 
-# Create your views here.
+@csrf_exempt
```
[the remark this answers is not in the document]
Leaversa: csrf exempt is still required because the cookie isnt properly configured in dev so views wont function without the flag. It can be removed in a future issue but we need this for now.

mesh/profiles/views.py
Leaversa: Issue already exists, I've edited the pr description to include it.

Issue #115, opened 2023-04-25
B4.2	HTTP Post Call Handling


GuyWhoCode:
Story: **As a User (Mentor, Mentee, or Educator), I want to be able to log in with my account so I can start using the service.**

Details: Handles the Post call to determine if the user account exists and matches the stored credentials using a boolean return value response. Might require collaboration with the person assigned to #72 to decrypt the password and check the database.

Technical Details:
- Password decryption
- Django
- HTTP REST Post
- MySQL

Issue #72, opened 2023-04-18
B1.3 Password Encryption


misslame:
Story: **As a User (Mentor, Mentee, or Educator), I want to be able to create an account and sign up with my email.**

Details: Write the encryption algorithm for passwords. 

Technical Details: 
- SHA-256
- Back-End
- Django
- Python
- Password Encryption

PR #220, opened 2023-06-23
Leaf/b1.3 password encryption

peppacaiou:
solves issues :
#115
#72  
updated an extra entry in the table for salts, because I can't append it to the encrypted password. 
inputs are passed In through the body with an x-www-form-urlencoded

endpoints: 
localhost/password/createAccount 
takes in an email, phone, number and password(isMentee/isMentor are optional and will default to false if not provided)
passwords are hashed with sha 256 , salted and then peppered
account creation with a phone number or email previously stored in the db will be denied access

stores the credential otherwise

localhost/password/checkAccount
takes in an email, phone number and password

verifies if a password is stored in the db.

image


misslame:
This is targeting Main, this needs to target `dev_new`

mesh/accounts/migrations/0002_account_salt.py
rezatnoMsirhC, line 1: Should these migrations files be version-controlled? I ask because they're auto-generated, and I'm not familiar with Django. I'm not sure if committing these adds any value.

mesh/accounts/models.py
```diff
@@ -3,8 +3,11 @@
 class Account(models.Model):
     accountID = models.AutoField(primary_key=True)
     email = models.EmailField()
-    encryptedPass = models.CharField(max_length=64)
-    
+
+    #please set editable to False after debugging process
+    encryptedPass = models.BinaryField(max_length=64,editable = True)
+    salt = models.BinaryField(max_length = 36,editable = True) 
+    ###
```
rezatnoMsirhC, line 10: What's this for?

mesh/accounts/views.py
```diff
@@ -1,2 +1,117 @@
...
+import bcrypt
+import os
+
+
+
+
+def encrypt(password ):
+    salt = bcrypt.gensalt(12)
+    
+    pepper = os.getenv("PEPPER")
```
rezatnoMsirhC, line 14: peppa

owwix (approved):
looks good

mesh/accounts/models.py
peppacaiou: editable is to be used for debugging in the admin console. I changed encryptedPass to a BinaryField to make storing and retrieving easier. Salt is stored alongside the password, but because I'm using Bcrypt, I don't really know any delimiters they don't use, so I figured it's more efficient to store it as a new field.

sosropartono (approved):
looks good to me

Issue #130, opened 2023-05-16
G1.6 Init Tags and Bridge Table


rezatnoMsirhC:
Write SQL queries to define the schema of and create the Tags and TagBridge tables.

This is the current schema for Tags:
- tagID (primary key)
- tagName
- isDefault

This is the current schema for TagBridge:
- accountID (primary/foreign key)
- tagID (primary/foreign key)

Refer to [the UML-Database diagrams.net Page](https://app.diagrams.net/#G18i2FGzLBHPOk-yKIdq-dVpT0CnBap6Ri) for the latest details.

PR #232, opened 2023-06-28
Leaf/1.6 tags bridge tables

sosropartono:
Original PR: #176 
Issue: #130


mesh/tags/models.py
```diff
@@ -1,9 +1,14 @@
...
-    pass
+    tagID = models.AutoField(primary_key=True)
+    tagName = models.CharField(max_length=255)
+    isDefault = models.BooleanField()
+
 class TagBridge(models.Model):
-    #TODO
-    pass
\ No newline at end of file
+    class Meta:
```
GuyWhoCode, line 11: I'm a little confused about the inclusion about the Meta class? Where is it being used? Do you really need it?

GuyWhoCode (changes requested):
Please look at my comment. Should be able to get re-review this PR quickly since I will also be working on my issue today.

mesh/profiles/urls.py
```diff
@@ -1,5 +1,6 @@
-from django.urls import path
+from django.urls import path, include
 from . import views as profile_views
+from django.http import HttpRequest
```
JRSiwiecki, line 3: Small nitpick, but is importing inlcude and HttpRequest necessary for this file? I can see that include might be necessary in the future due to some other urls.py files importing it as well, but I don't see how HttpRequest is necessary.

mesh/tags/models.py
JRSiwiecki: @GuyWhoCode From what I can tell, [the Meta class](https://docs.djangoproject.com/en/4.2/ref/models/options/) just represents metadata for models, so while it doesn't seem explicitly necessary, it does seem that it can be useful purely for information. as something like unique_together is a metadata field provided by the Meta class. Although it might look a little more clear in code if the class Meta on line 11-12 was placed after the tagID and accountID definitions.

JRSiwiecki (approved):
Since the other two aforementioned details seem minor, this looks good to me!

mesh/profiles/urls.py
sosropartono: I think this was merged from another branch, but your reasoning makes sense, will remove

K0725 (commented):
looks good to me

GuyWhoCode (approved):
Looks good, thanks for clarifying my comment!

Issue #131, opened 2023-05-16
G1.7 Init Notifications Table


rezatnoMsirhC:
Write a SQL query to define the schema of and create the Notifications table.

This is the current schema:
- notificationID (primary key)
- accountID (foreign key)
- notificationContextJson
- timestamp

Refer to [the UML-Database diagrams.net Page](https://app.diagrams.net/#G18i2FGzLBHPOk-yKIdq-dVpT0CnBap6Ri) for the latest details.

PR #233, opened 2023-06-28
Leaf/1.7 notifications table

sosropartono:
Original: #178 
Issue: #131


GuyWhoCode (approved):
Database schema matches the UML diagram. Nice work!

mesh/notifications/models.py
```diff
@@ -1,7 +1,11 @@
...
 
 class Notification(models.Model):
-    #TODO
-    pass
\ No newline at end of file
+    accountID = models.ForeignKey(Account, on_delete=models.CASCADE)
+    notificationID = models.AutoField(primary_key=True)
+    timeStamp = models.TimeField()
+    notificationContextJSON = models.JSONField()
+
```
JRSiwiecki: Should this table also have a Meta class for metadata or is it unnecessary, otherwise looks good to me!

misslame:
@sosropartono I am reverting this. This has commits that are in Main ONLY. This means you had branched off of main, NOT dev_new

@GuyWhoCode & @JRSiwiecki  you need to look at the commits. There should be no reason why commits made from me would be in his work. You both approved which holds you also responsible for when *bad* commits make it into dev.

Issue #121, opened 2023-05-02
F1.5 Match Figma for Dark Theme


Leaversa:
Story: As a night owl account owner mentor, mentee, or educator, I want to be able to switch to a dark theme version of the site so I can look at it better.

Details: Fill in the gaps from the figma design after implementing the theme. See if there are any discrepancies that needs to be specifically applied or fixed. 

Technical Details:
- Front-end
- UI/UX
- React
- CSS 
- MUI

PR #235, opened 2023-06-29
Leaf/f1.5 match figma for dark theme

Darren52BT:
Addresses #121.

I have created a dark and light MUI theme for the application, which can be found in the theme folder. I have basically created new colors and overrode some basic color variants for the Palette, and I've created styles for specific components in specific states. I've applied to the main application in the index file, and I've adapted the custom footer, login form, login page, logged out home page, the profile cards, and the swiping card page to use the theme. Some of these pages already had existing themes for things like font size, so I basically merged the main theme and these themes together. I have stacked all of these components on top of each other in the index file, so upon running the application you should be able to scroll down and see these components styled according to the theme. If you want to change the theme to light or dark, just go into the MainTheme file, scroll to the bottom, and change the mode argument to light or dark. This is a scuffed way of doing it but my main concern is just getting the theme done.
image

**Concerns:** 

I did take liberties with naming the custom colors that were defined in the figma. I used module augmentation in the theme.d.ts file to introduce these new names. I tried to name them accordingly to how they are being used, but it is not perfect so if anyone has any better names for them I would greatly appreciate them. 

I also noticed some inconsistencies in the figma design There are differences in what should be the icon/active, text/primary, and text/secondary colors between the pages like Profile View/Swiping Carousel and Sign Up/Logged In Home. I think this might be due to the fact that none of the dark theme colors in the figma are being mapped to any specific variants like the light theme is, so maybe there were some mistakes being made there. The colors look very similar to each other, so I just chose one version to use for everything.

 I also didn't add two colors for certain elements in the figma. One color (#D9D9D9) is for the avatar background/border color in the profile cards/profile view and the image container color in the Logged Out Home Page. The other (#F2E8DE) is being used for the logo and links for the headers on every page. These colors are the same across both dark and light mode. I haven't added them yet because I dont know if its necessary to add them in the themes if they are just constant between both, and because the first color is being used in a lot of areas so Im not sure what to call it.

image
image
image
image
image
image

### DIFF
--- a/meshapp/src/theme/MainTheme.tsx
+++ b/meshapp/src/theme/MainTheme.tsx
@@ -222,6 +222,7 @@ function getComponents(mode: PaletteMode) {
   };
 }
 
+//change arg to light or dark, depending on which mode you want, to switch to that mode
 const theme = createTheme({
   ...getDesignTokens("dark"),
 });
--- a/meshapp/src/theme/MainTheme.tsx
+++ b/meshapp/src/theme/MainTheme.tsx
@@ -1,9 +1,22 @@
 import { PaletteMode, createTheme } from "@mui/material";
 import React from "react";
 
+//gets pallete colors and component colors based on mode
 const getDesignTokens = (mode: PaletteMode) => ({
   palette: {
     mode,
+    ...getPalette(mode),
+  },
+
+  components: {
+    ...getComponents(mode),
+  },
+});
+
+//returns custom palette colors based on dark or light mode
+function getPalette(mode: PaletteMode) {
+  //light mode
+  return {
     ...(mode === "light"
       ? {
           primary: {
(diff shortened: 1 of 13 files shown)


meshapp/src/theme/MainTheme.tsx
```diff
@@ -0,0 +1,229 @@
...
+          },
+        },
+      },
+    },
+  };
+}
+
+const theme = createTheme({
+  ...getDesignTokens("dark"),
+});
```
PattyCakesxD, line 227: Include comment here to indicate you can swap b/w light and dark mode (for future tasks that works with it)

PattyCakesxD (approved):
Should the cocogoose font also be included in the themeing?

Rest lgtm, nice work!

Darren52BT:
> Should the cocogoose font also be included in the themeing?
> 
> Rest lgtm, nice work!

Well the cocogoose font looks kinda weird and doesn't quite match the figma so I just avoid using it for now. It's simple enough to add it later anyways

Leaversa (approved):
lgtm!

Issue #74, opened 2023-04-18
B1.5 Email Confirmation


misslame:
Story: **As a User (Mentor, Mentee, or Educator), I want to be able to create an account and sign up with my email.**

Details: Send an email to the user after they sign up and ensure they are verified. Include any error handling and UX/UI messaging that suggests they need to confirm their email. You may need to do GET handling, modify the database, etc. This is a fairly large task but all aspects should be done together

Technical Details: 
- Full-Stack
- Django
- Email Verification
- React
- HTTP REST GET
- SQL

misslame:
Roll over

Issue #218, opened 2023-06-17
Add user token field to Settings table


JRSiwiecki:
For issue [#74](https://github.com/LetsMesh/Site/issues/74), one current plan for email confirmation is to store a generated user token in the settings table for the user, and when an email is sent to them, we check that the token in the url is the same one stored in the settings table, so that other users can't verify someone's email for them in a sense. 

If implemented, this would require a change in issue [#126](https://github.com/LetsMesh/Site/issues/126) where the settings table is initialized, and likely also an update for the database UML.

Related pull requests: PR [#237](https://github.com/LetsMesh/Site/pull/237), PR [#197](https://github.com/LetsMesh/Site/pull/197)

JRSiwiecki:
For clarification, the token is a hex-string of 32 characters.

JosephKahale:
@JRSiwiecki Just wanted to make sure this is still the plan, so I can change the UML as well as the settings table.

JRSiwiecki:
@JosephKahale Yes this is still the idea I'd like to work with, thanks!

JosephKahale:
@JRSiwiecki Would you still use the isVerified bool in the settings table? Or is it not needed anymore due to the 32 char token?

JRSiwiecki:
@JosephKahale I figure that it would still be necessary so as to not send repeat confirmation emails once a user is already verified, but if you'd like to remove the field entirely then I suppose something could be done where if the user's token field isn't null, then in theory we wouldn't need isVerified, but I feel as though it would be simpler to just keep the isVerified field.

JRSiwiecki:
Closed with [#245 ](https://github.com/LetsMesh/Site/pull/245/files)

PR #237, opened 2023-06-29
Leaf/b1.5 email confirmation

JRSiwiecki:
Issue: [#74](https://github.com/LetsMesh/Site/issues/74)

When a user visits an email confirmation url path, an email is sent to the user, and a unique token is generated and sent as a url to the user, and saved in the settings table. If the saved token and the token in the url matches, then the user's email is confirmed.


JRSiwiecki:
Likely needs to be updated when issue [#218](https://github.com/LetsMesh/Site/issues/218) is closed, as currently the email confirmation process does not use the database as it should.

mesh/confirmation/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path, include
+from . import views as confirmation_views
+
+urlpatterns = [
+    path('<str:user_email>/', confirmation_views.email_confirmation, name="email_confirmation"),
+    path('<str:user_email>/<str:token>/', confirmation_views.confirm_token, name="confirm_token"),
```
sosropartono, line 6: Should we have a path specifically for emails? Like /emails/<str...

sosropartono (approved):
One question, otherwise LGTM

mesh/confirmation/urls.py
JRSiwiecki: I figure that the confirmation path in the urls.py in the root mesh folder where it is already confirmation/<str:user_email>/ handles this idea? Unless you mean add the /email/ on top of the confirmation path for clarity.

JosephKahale (approved):
LGTM

PR #222, opened 2023-06-24
Leaf/b1.2_http_post_call_handling_new

bluesimp1102:
Most of the changes are pretty much the same as the last PR, I added some new stuff so the account module can be added correctly to the new main API module with the refactored structures.

Issue https://github.com/LetsMesh/Site/issues/67

Issue #67, opened 2023-04-18
B1.2 HTTP Post Call Handling


misslame:
Story: **As a User (Mentor, Mentee, or Educator), I want to be able to create an account and sign up with my email.**

Details: Handle what happens when a post call is sent after creating an account. 

Technical Details:
- HTTP Rest Post call handling
- Back-end
- Django
- Sql
- Databases

PR #238, opened 2023-06-29
Leaf/b1.2_http_post_call_handling

bluesimp1102:
Issue: https://github.com/LetsMesh/Site/issues/67
Original PR: https://github.com/LetsMesh/Site/pull/222
Creating a new PR because the old branch was from main instead of dev_new


mesh/accounts/views.py
```diff
@@ -1 +1,105 @@
...
+            phoneNum = data['phoneNum'], 
+            displayTheme = data['displayTheme'], 
+            enabled2Factor = data['enabled2Factor'],
+            isMentor = data['isMentor'],
+            isMentee = data['isMentee']
+        )
+        return JsonResponse({'account_id': account.id}, status=201)
+
+
+@method_decorator(csrf_exempt, name='dispatch')
```
misslame, line 55: Why make it csrf exempt

bluesimp1102: I removed it. The initial thought was having it there to having the protection but I didn't realize we won't have it if I put it there.

owwix (changes requested):
remove use of csrf exempt decorators, the csrf middleware is temporarily disabled.

GuyWhoCode (approved):
Nice!

owwix (approved):

Issue #64, opened 2023-04-18
B1.1 UI/UX Display for sign up


misslame:
Story: **As a User (Mentor, Mentee, or Educator), I want to be able to create an account and sign up with my email**

Details: User Interface including all the functionality such as password protection on the display, clicking buttons, info tooltips, etc.  (Follow the Figma Design) 

Technical Details:
- Front-end
- React
- HTML/CSS
- UI/UX

misslame:
@jamsge Please Work on Sign UP Step 1
@ethan-ngo  Please work on Sign Up Step 2
@Shadowowl888  Please work on Sign UP Step 3

jamsge:
https://github.com/LetsMesh/Site/tree/leaf/B1.1-UIForSignUp

@ethan-ngo @Shadowowl888

PR #250, opened 2023-07-08
B1.5 Email Confirmation 

### LINKED ISSUE #74: B1.5 Email Confirmation

JRSiwiecki:
Resolves issue #74 

### FEATURE
- After a user signs up, they should be redirected to `.../confirmation/<email>/,` at which point a unique 8 bit token along with the current time as an integer will be created and stored in the Settings table for their account, and a message will be displayed saying to check their email.
- A pretty email will be sent to the user containing a welcoming message and a link to actually verify their email. and when they follow it, their email will be verified and the Settings table will be updated accordingly. 
- If a user tries to request email verification again in a 10 minute window, they will be told to wait until 10 minutes have elapsed since their last attempt to verify their email.
- If a user is already verified, they will be notified and no new token will be generated.
- If a user does not exist and they attempt verification, they will be notified that the account does not exist.

### TO-DO
- Currently uses "personal" emails, which will get your account locked if sent to someone NOT on your current IP. This should likely be solved by setting up the project to use some free email sending service like Sendgrid, as outlined in issue #251.
- The pages for displaying to a user that an email has been sent to them and that their email has been verified would likely be completed with issue #64.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 # Django
 *.pyc
 *.sqlite3
+media
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@
 # Django
 *.pyc
 *.sqlite3
-media
--- a/mesh/confirmation/views.py
+++ b/mesh/confirmation/views.py
@@ -30,11 +30,14 @@ def email_confirmation(request, user_email):
     # Prevent user from generating too many tokens/receiving too many emails
     settings_data = get_settings_data(account_id)
     
+    EMAIL_CONFIRMATION_COOLDOWN_SECONDS = 600
+    EMAIL_CONFIRMATION_COOLDOWN_MINUTES = EMAIL_CONFIRMATION_COOLDOWN_SECONDS // 60
+    
     seconds_since_last_token = calculate_timestamp_difference(settings_data["verificationToken"])
     minutes_since_last_token = seconds_since_last_token // 60
     
-    if seconds_since_last_token <= 600:
-        return HttpResponse(user_email + ": It has been " + str(minutes_since_last_token) + " minutes since last attempt at verifying email! Please try again in " + str(10 - minutes_since_last_token) + " minutes.")
+    if seconds_since_last_token <= EMAIL_CONFIRMATION_COOLDOWN_SECONDS:
+        return HttpResponse(user_email + ": It has been " + str(minutes_since_last_token) + " minutes since last attempt at verifying email! Please try again in " + str(EMAIL_CONFIRMATION_COOLDOWN_MINUTES - minutes_since_last_token) + " minutes.")
     
     # Generate a token for the user
     verification_token = generate_token_with_timestamp()
--- a/mesh/confirmation/views.py
+++ b/mesh/confirmation/views.py
@@ -1,0 +1,37 @@
+from django.http import HttpResponse
+from django.core.mail import send_mail
+from django.template.loader import render_to_string
+from django.utils.html import strip_tags
+import secrets
+
+# TODO: It would likely be easier to just store the 
+# unique token generated for the user in their settings,
+# and then when they click the link sent to them we can verify
+# that the one sent to them is the same as the one stored
+# in their settings.
+
+# Generate random token for user and send confirmation email
+def email_confirmation(request, user_email):
+    # Generate a token for the user
+    confirmation_token = secrets.token_hex(16)
+
+    # Construct the confirmation URL
+    confirmation_url = request.build_absolute_uri(f"/confirmation/{user_email}/{confirmation_token}/")
+
+    # Render the email template with the confirmation URL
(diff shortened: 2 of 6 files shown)


mesh/settings.py
```diff
@@ -12,6 +12,9 @@
 
 from pathlib import Path
 import os
+from dotenv import load_dotenv
+
+load_dotenv()
```
GuyWhoCode, line 17: Is there a reason you used `dotenv`? In this file, the `os` library is used to import the `.env` variables, so this is redundant from my perspective.

.gitignore
```diff
@@ -6,3 +6,4 @@
 # Django
 *.pyc
 *.sqlite3
+media
```
GuyWhoCode, line 9: What's a media file?

mesh/confirmation/views.py
```diff
@@ -0,0 +1,182 @@
...
+    if email_already_verified:
+        return HttpResponse(user_email + ": User email already verified!")
+
+    # Prevent user from generating too many tokens/receiving too many emails
+    settings_data = get_settings_data(account_id)
+    
+    seconds_since_last_token = calculate_timestamp_difference(settings_data["verificationToken"])
+    minutes_since_last_token = seconds_since_last_token // 60
+    
+    if seconds_since_last_token <= 600:
```
GuyWhoCode, line 39: Use constants! Replace 600 with something like `SECONDS_COOLDOWN = 600`. Try to avoid having floating numbers in your code if it's meaningful.

mesh/confirmation/views.py
```diff
@@ -0,0 +1,182 @@
...
+    account_id = account_data["accountID"]
+
+    return account_id
+
+
+# Grab settings data by the account_id
+# settings_data is a python dict containing
+# data for each field in the Settings model
+def get_settings_data(account_id):
+    raw_settings_data = Settings.objects.filter(accountID=account_id).values()
```
GuyWhoCode, line 129: For efficiency purposes, it might be better to use `Settings.objects.get` since accountID is unique, meaning that there should be only one entry that matches the query.
[Query Get Documentation](https://docs.djangoproject.com/en/4.2/topics/db/queries/#retrieving-a-single-object-with-get)

GuyWhoCode (changes requested):
Please see comments on some minor adjustments that I recommend. When you get the changes done, please ping me again in Discord.

.gitignore
JRSiwiecki: It seems to just hold locally stored data such as images. When I was testing email confirmation along with using a user's preferred names in the email, I needed to create dummy data and doing so required me to give the data an image, which then created a media folder, with an image folder inside it, holding the image folder that I placed.

Basically this is just to ignore those files.

mesh/settings.py
JRSiwiecki: I needed this for some reason at some point, but I just tested it again and it worked fine without it, so I'll remove this!

mesh/confirmation/views.py
JRSiwiecki: Good point! Will fix.

mesh/confirmation/views.py
JRSiwiecki: This is another good point that I will fix, as not only is it more efficient, the way I was doing it before required you to use the python dictionary subscripts to access parts of the object, whereas getting the object itself allows for autocomplete and generally more clear code.

GuyWhoCode (approved):
I would remove the media from the `.gitignore` just to remove anything that shouldn't be there. But other than that really small thing, everything looks good!

JRSiwiecki:
> I would remove the media from the `.gitignore` just to remove anything that shouldn't be there. But other than that really small thing, everything looks good!

My main problem with this is that if I remove media from the .gitignore, then I am just committing a random .jpg image that is not necessary for anyone other than my own data.

GuyWhoCode:
> > I would remove the media from the `.gitignore` just to remove anything that shouldn't be there. But other than that really small thing, everything looks good!
> 
> My main problem with this is that if I remove media from the .gitignore, then I am just committing a random .jpg image that is not necessary for anyone other than my own data.

That can be solved by choosing what gets added to the commit either in the VS Code interface or using `git add`. You can use `git status` to see all changed files and `git add` to choose which files get committed. It's more inconvenient than just adding the folder to the `.gitignore` but making any changes to that file gets uploaded to GitHub (unless you choose not to)

sosropartono (approved):
Looks good to me!

Issue #134, opened 2023-05-20
F3.1 Logged In Home Screen


misslame:
Story: **As a product owner, I want to be able to ensure a safe and welcoming environment on the site/app by restricting language to only include appropriate language.**

Details: Follow the figma design of the home screen that is logged in. Create any issues for gaps in functionality and describe their requirements properly following task format.

Technical Details:
- Front-end
- React 
- MUI
- UI/UX

rllov:
https://github.com/LetsMesh/Site/tree/leaf/F3.1-Logged_In_Home_Screen

PR #253, opened 2023-07-12
Leaf/f3.1 logged in home screen

tdang2180:
#134 
Created the logged-in home screen
- Created a message footer
- Fixed Header theme


meshapp/src/components/Header.tsx
```diff
@@ -7,11 +7,15 @@ import React from "react";
...
+    <AppBar
+      position="static"
+      sx={{ backgroundColor: "primary.main" }}
+      elevation={0}
+    >
       <Toolbar>
         {/*---Left Side Menu Elements (Menu/Notification Icon)---*/}
         <Grid container>
-          <IconButton sx={{ mr: 2, color: "#0000008F" }}>
+          <IconButton sx={{ mr: 2, color: "secondary.main" }}>
```
Darren52BT, line 18: the color should be "action.active", you can see the difference if you switch to light mode

meshapp/src/components/Header.tsx
```diff
@@ -20,7 +24,7 @@ const Header = () => {
             edge="start"
             color="inherit"
             aria-label="menu"
-            sx={{ mr: 2, color: "#0000008F" }}
+            sx={{ mr: 2, color: "secondary.main" }}
```
Darren52BT, line 27: same issue as the other icon button

Darren52BT (changes requested):
Aside from the button color, everything looks good for the most part. There is also the issue of all the sizes becoming way too big after we go below 600px. I dont think we're too concerned about styles if the screen width is super small, but I would suggest maybe extending those sizes for a minimum screen width of 600px to screen widths below 600px as well. It might start breaking down again at around 400px but at that point I don't think that's a big deal, we can go back to fix it if we need to.

meshapp/src/components/Header.tsx
tdang2180: 👍

Darren52BT (approved):
lgtm

Issue #21, opened 2023-03-12
A4.8 HTTP Post Call Handling


misslame:
﻿﻿Story: ﻿﻿﻿As a User (Mentor, Mentee, or Educator), I want to be able to select tags from a catalog to include on my profile that indicates my professional and casual interests.﻿
﻿
﻿﻿Details: ﻿
﻿Receive POST Calls in order to store new tags to the database for the specified User.
﻿
﻿﻿Technical Information:
- ﻿﻿Django
- POST Api Call
- Error Handling & Response
- Back-end

PR #256, opened 2023-07-14
Leaf/a4.8 http post call handling

K0725:
Task Post call Handling for interest tag #21 

- Implemented HTTP Post call handling for tag field.


JRSiwiecki (changes requested):
Your code looks mostly fine, however it seems that your branch has changes from other users that weren’t assigned this issue, which should be fixed, as per the announcement on 06/26/23 in #sprint-announcements on the discord.

This could be fixed in a few ways, but the simplest way could be creating a new branch off of dev_new, git cherry-picking your commits only, deleting this current branch, and then requesting a new pull request.

mesh/tags/views.py
```diff
@@ -35,3 +35,35 @@ def user_tags(request):
...
+    if request.method == "POST":
+        data = request.POST
+        required_fields = ["tagName", "isDefault", "accountID"]  
+        if not all(field in data for field in required_fields):
+            response_data = {'status': "Error: Missing field"}
+            return JsonResponse(response_data, status=400)
+        
+        try:
+            # Creating new tag object
+            tag = Tag.objects.create(
```
GuyWhoCode, line 50: You're missing an import here

mesh/tags/views.py
```diff
@@ -35,3 +35,35 @@ def user_tags(request):
...
         return JsonResponse(response_data, status=200)
+
+
+def post_tags(self, request):
+    if request.method == "POST":
+        data = request.POST
+        required_fields = ["tagName", "isDefault", "accountID"]  
+        if not all(field in data for field in required_fields):
+            response_data = {'status': "Error: Missing field"}
+            return JsonResponse(response_data, status=400)
```
GuyWhoCode, line 46: Have you tested your code with Postman? When I tried making a post request, I'm unable to get a response when there are no fields.

GuyWhoCode (changes requested):
Please re-request my review after you've taken a look at my comments. Also, change `settings.py` to not include an import to the `dotenv` library,

mesh/tags/views.py
K0725: ohh I see

mesh/tags/views.py
K0725: I've tested the postman but I just created my own test cases. So I will get back to on this one

Issue #120, opened 2023-05-02
F1.3 HTTP Get Call Handling


Leaversa:
Story: As a night owl account owner mentor, mentee, or educator, I want to be able to switch to a dark theme version of the site so I can look at it better.

Details: GET call handling for account display theme settings 

Technical Details:
- Back-end
- MySQL
- Django
- Python
- HTTP REST Get

PR #260, opened 2023-07-17
Leaf/f1.3 http get call handling

shotoh:
issue #120


mesh/tests/settings_tests.py
```diff
@@ -0,0 +1,46 @@
+import json
```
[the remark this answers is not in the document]
misslame: Hmm maybe we can keep it at "settings_test" if the modal is also called "Setting"

mesh/accountSettings/views.py
```diff
@@ -1,3 +1,30 @@
...
+        response = {
+            "status": "success"
+        }
+        if data.get("accountID") is None:
+            response.update({"status": "error"})
+            response.update({"message": "Missing account ID."})
+            return JsonResponse(response)
+        else:
+            if Settings.objects.filter(accountID=int(data.get("accountID"))).exists():
+                settings = Settings.objects.filter(accountID=int(data.get("accountID")))
```
JRSiwiecki, line 18: This is somewhat expensive to call twice, no? I think this approach might make more sense rather than calling the same method twice:

settings = Settings.objects.filter(accountID=int(data.get("accountID")))

if settings:

mesh/tests/settings_tests.py
JRSiwiecki, line 1: Thank you for making test cases :)

JRSiwiecki (changes requested):
Please review my comments, and then ping me on discord for approval! Good work 👍🏼

JRSiwiecki (approved):
Good work!

misslame (approved):

Issue #108, opened 2023-04-25
B2.1 Forgot Password Link on Login page


GuyWhoCode:
Story: **As a User (Mentor, Mentee, or Educator), I want to be able to request a password change if I forgot my password.**

Details: Changes the login page to allow for email text only. This will remove or hide the password field, showing only the email field on the login page. When the form is submitted, the form will send a post request for a password change. 

Technical Details:
- React
- Full-stack
- HTTP REST Post
- Axios
- UI/UX

PR #261, opened 2023-07-20
Leaf/b2.1 forgot password link on login page

owwix:
B2.1 Forgot Password Link on Login page #108

**Home page:** 
Screen Shot 2023-07-28 at 10 31 39 AM

**Home page after clicking Forgot Password button:**
Screen Shot 2023-07-28 at 10 32 08 AM


meshapp/src/home/login.tsx
```diff
@@ -52,6 +53,18 @@ const SignUp = () => {
...
+  const [showForgotPassword, setShowForgotPassword] = useState(false);
+
+  const updateShowForgotPasswordState = () => {
+    if (showForgotPassword == false) {
+      setShowForgotPassword(true)
+    }
+    if (showForgotPassword == true) {
+      setShowForgotPassword(false)
+    }
+
```
rezatnoMsirhC, line 66: 1. This should be done without if-statements. 
2. If you really want to keep the if-statements, this should be if/else. `showForgotPassword` cannot be both true and false. An interviewer probably wouldn't be happy if they saw this. 

https://softwareengineering.stackexchange.com/questions/12807/make-a-big-deal-out-of-true

owwix: Just addressed this, thank you for that!

meshapp/src/home/login.tsx
```diff
@@ -52,6 +53,13 @@ const SignUp = () => {
 };
 
 const Login = () => {
+
+  const [showForgotPassword, setShowForgotPassword] = useState(false);
+
+  const updateShowForgotPasswordState = () => {
+    setShowForgotPassword(prevState => !prevState)
```
rezatnoMsirhC, line 60: Wasn't this based on `showForgotPassword` previously? Should it be based on `prevState`?

rezatnoMsirhC: Oh nevermind I think I'm looking at the wrong section 😅

meshapp/src/home/logged-out/home.tsx
```diff
@@ -1,16 +1,29 @@
...
+  const [showForgotPassword, setShowForgotPassword] = useState(false);
+
+  const updateShowForgotPasswordState = () => {
+    if (showForgotPassword == false) {
+      setShowForgotPassword(true)
+    }
+    if (showForgotPassword == true) {
+      setShowForgotPassword(false)
+    }
+  }
```
rezatnoMsirhC, line 25: Looks like I forgot to comment the same thing here. Why was this code duplicated though? Can't the logic be moved out into a common function both files can use?

rezatnoMsirhC:
Please provide some testing info/screenshots

meshapp/src/home/login.tsx
owwix: https://stackoverflow.com/questions/73993973/understanding-prevstate-in-react

meshapp/src/home/logged-out/home.tsx
owwix: https://github.com/LetsMesh/Site/issues/269 created an issue to track this

rezatnoMsirhC (approved):

meshapp/src/home/logged-out/home.tsx
```diff
@@ -59,7 +67,7 @@ export default function LoggedOutPage() {
               filter: "drop-shadow(4px 4px 3px rgba(0,0,0,.6))",
             }}
           >
-            {LoginInput()}
+            {showForgotPassword ? <ForgotPassword updateShowForgotPasswordState={updateShowForgotPasswordState} /> : <LoginInput updateShowForgotPasswordState={updateShowForgotPasswordState} />}
```
misslame, line 70: This seems repetitive.  Is there anyway to incorporate it into login?

misslame (changes requested):
Is it possible to have forgot password logic & rendering be a part of the "login" component rather than the logic scattered about on the login screen & home screen?

meshapp/src/home/logged-out/home.tsx
owwix: Refactored :)

misslame (approved):
:D looks good

PR #263, opened 2023-07-21
Leaf/b1.1 UI/ux display for sign up

Darren52BT:
I've created three form steps for the sign up page. This is all being rendered on one page. Clicking on the continue button will trigger validation for the current visible inputs, and if there are no errors it will move on to the next step. If there are errors, it will display them.

Here are the following inputs for each step and the validation rules. Im not sure what else to put for validation rules beyond these, so if anyone has anything they want to add feel free to suggest them:

 Step One
-------------
- First Name (single line text field) : required
- Last Name (single line text field): required
- Nickname (single line text field): no rules
- Phone Number (single line text field) : required, fits this regex 
     /^(\+\d{1,2}\s)?\(?\d{3}\)?[\s.-]\d{3}[\s.-]\d{4}$/
- Country (select): required
- State (select): required, state must belong to the current selected country
- Email Address (single line text field): required, fits this regex   
/^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
- Accept Terms and Conditions (checkbox): required
- Sign Up for Email Updates(checkbox): no rules

Step Two
-------------
no inputs, though the inputted email from the previous step is displayed

Step Three
---------------
- Name (single line text field) : required
- Location (single line text field): required
- Title (single line text field): no rules
- Pronouns( single line text field): no rules
- User Type (select): required, value must match Mentee, Mentor, Educator
- Interests (Autocomplete): required
- Label: (multiline text field): required
- Picture (custom file input): has to be image file

I've created custom reusable input components integrated with React-Hook-Form's controller, which is being used everywhere except for the picture input, which I mainly copied from Patrick's work on the profile page with some small edits to set values on the form. The custom select component in particular takes in a callback, which I implemented in order for the country and the state select options to be tied to each other.

For displaying errors, I've made use of an error tooltip that will appear at the start of the corresponding input element. Hovering over the tooltip will display the error messages. Im not sure about the current messages being displayed so if anyone has any better ones feel free to suggest them.

In general, the inputs will be on blur. There are some exceptions for the autocomplete, select, and checkbox inputs, which will validate on change. I could have made it to be on change for the other inputs, but that requires higher performance from the browser since they will be validating on every keystroke, so I made the decision not to do that.

Step Displays:
----------------------------------
image
image

image
image
image

image

image

If you want to see the current form data (or errors) after attempting to submit, just take a look in the dev tools console
image


misslame (approved):
Pretty self explanatory, well broken down into methods IMO, loving it.

misslame:
PERFECT PR

misslame:
I LOVE THIS ONE

meshapp/src/components/SignUp/SignUp.tsx
```diff
@@ -0,0 +1,225 @@
...
+import Step from "@mui/material/Step";
+import StepLabel from "@mui/material/StepLabel";
+import { useState } from "react";
+import Grid from "@mui/material/Grid";
+import Button from "@mui/material/Button";
+import { useForm, SubmitHandler, FormProvider } from "react-hook-form";
+import ChevronLeftIcon from "@mui/icons-material/ChevronLeft";
+import ArrowForwardIcon from "@mui/icons-material/ArrowForward";
+import { Theme, ThemeProvider, Typography, createTheme } from "@mui/material";
+const steps = ["Create Account", "Verify Email", "Go to Profile"];
```
rezatnoMsirhC, line 14: I think this should be an enum instead

Example:
```
enum Test {
  A = "a",
  B = "b",
  C = "c"
}

console.log(Object.keys(Test)); // [ 'A', 'B', 'C' ]
console.log(Object.values(Test)); // [ 'a', 'b', 'c' ]
console.log(Test.A); // a
```

meshapp/src/components/SignUp/StepTwo.tsx
```diff
@@ -0,0 +1,275 @@
...
+  p: 4,
+  display: "flex",
+  justifyContent: "center",
+  alignItems: "center",
+  bgcolor: "#D9D9D9",
+  color: "#27383A",
+  textAlign: "center",
+};
+
+//TODO (#262): rework send verification buttons to actually send verification emails to the input email, and for the success email verification modal to appear upon actual verification
```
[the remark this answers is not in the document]
Darren52BT: 🗡️

meshapp/src/components/SignUp/SignUp.tsx
```diff
@@ -0,0 +1,225 @@
...
+    const continueToNext = () =>
+      setActiveStep((prevActiveStep) =>
+        prevActiveStep < steps.length ? prevActiveStep + 1 : prevActiveStep
+      );
+
+    let isValid = false;
+
+    // trigger validation for each step unless it is 2nd step (where there aren't any inputs)
+    //if we're on the last step, then trigger submit
+    if (activeStep === 0) {
```
rezatnoMsirhC, line 101: This is one place where having the steps as an enum would help. This could be `if (activeStep === Step.CREATE_ACCOUNT)` or something similar.

meshapp/src/components/SignUp/SignUp.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+  //handles the submission of the data in the current step
+  const handleContinue = async () => {
+    //function for continuing to next step
+    const continueToNext = () =>
+      setActiveStep((prevActiveStep) =>
+        prevActiveStep < steps.length ? prevActiveStep + 1 : prevActiveStep
+      );
+
+    let isValid = false;
```
rezatnoMsirhC, line 97: Is what valid? I think this should be more descriptive.

meshapp/src/components/SignUp/StepOne.tsx
```diff
@@ -0,0 +1,380 @@
...
+function PhoneNumber() {
+  return (
+    <CustomStandardTextField
+      label="Phone Number"
+      id="phone-number"
+      fieldName="phoneNumber"
+      validators={{
+        pattern: (value) => {
+          let phoneNumRegex =
+            /^(\+\d{1,2}\s)?\(?\d{3}\)?[\s.-]\d{3}[\s.-]\d{4}$/;
```
rezatnoMsirhC, line 243: Store this and other string variables as constants somewhere else instead? I don't think we need to allocate these each time, though I'm not sure how TypeScript handles these (it might optimize by making them constants behind the scenes).

meshapp/src/components/SignUp/SignUp.tsx
Darren52BT: So the activeStep variable is being used to both reference the index of the current form component in stepComponents and which label we are using in the stepper. Currently I just added in an enum, but I've tied each enum member to their corresponding index instead so that activeStep can still be used to both index the form component and be used to referenced to the enum. The conditionals for validation triggering now reflect what you suggested, but I'm still using the original step array (renamed to stepLabels) for the stepper to display.

meshapp/src/components/SignUp/StepOne.tsx
Darren52BT: does it matter whether Im storing these strings at the top of StepOne or if I just put them at the top level of these component functions (like for example putting the phone regex right above the return statement of PhoneNumber())?

meshapp/src/components/SignUp/StepOne.tsx
```diff
@@ -0,0 +1,374 @@
...
+//password text field
+function Password() {
+  return (
+    <CustomPasswordTextField
+      fieldName="password"
+      id="password"
+      label="Password *"
+      validators={{
+        minLength: (value) => {
+          return value.length > 8 || "Password length needs to be at least 8.";
```
JRSiwiecki, line 278: Do you mean for this to be value.length >= 8? If someone enters a password of length 8, it still shows this error, so either the message needs to be changed to say it must be at least length 9, or the condition must be changed.

meshapp/src/components/SignUp/StepOne.tsx
```diff
@@ -0,0 +1,374 @@
...
+  return (
+    <CustomStandardTextField
+      label="Phone Number"
+      id="phone-number"
+      fieldName="phoneNumber"
+      validators={{
+        pattern: (value) => {
+          let phoneNumRegex =
+            /^(\+\d{1,2}\s)?\(?\d{3}\)?[\s.-]\d{3}[\s.-]\d{4}$/;
+          return phoneNumRegex.test(value) || "Invalid phone number format";
```
JRSiwiecki, line 244: It would be nice if this error message would explain the format you need, otherwise it is up to the user to guess.

meshapp/src/components/SignUp/inputs/CustomAutoComplete.tsx
```diff
@@ -0,0 +1,95 @@
...
+  InputAdornment,
+  TextField,
+  Tooltip,
+} from "@mui/material";
+import { Controller, useFormContext } from "react-hook-form";
+import { IFormInput } from "../SignUp";
+import ErrorIcon from "@mui/icons-material/Error";
+
+//takes in boolean stating whether its required or not,
+//strings for lable and id,
```
JRSiwiecki, line 13: Small spelling error here! 🤓🤓🤓

meshapp/src/components/SignUp/inputs/CustomSelect.tsx
```diff
@@ -0,0 +1,91 @@
...
+            required: args.required ? "This is required" : false,
+            validate: args.validators,
+          }}
+          render={({ field: { onChange, value } }) => (
+            <Select
+              displayEmpty
+              notched={true}
+              value={value}
+              onChange={(event) => {
+                onChange(event);
```
JRSiwiecki, line 46: When I run your branch myself, I get an error on these lines, specifically with onChange(event); so it might be worth checking if there are some dependency/package issues here.

JRSiwiecki (changes requested):
Please review my comments, and I'll approve after! Good work though! 👍🏼

meshapp/src/components/SignUp/inputs/CustomSelect.tsx
Darren52BT: There appears to be no problems when React-Hook-Form is on version 7.44.2. Upon making the PR I merged the two differing package.json together, and I chose the higher version 7.45.1, which results in this error happening. Im not sure what to do other than changing the version back.

Darren52BT: actually nvm i fixed it

JRSiwiecki: Good find :) 🥇

JRSiwiecki (approved):
Good work! 🌮

rezatnoMsirhC (approved):

PR #267, opened 2023-07-28
Leaf/a1.6 profile picture http get call handling

shotoh:
issue #6

### DIFF
--- a/mesh/profiles/views.py
+++ b/mesh/profiles/views.py
@@ -5,7 +5,6 @@ from django.http import JsonResponse
 from mesh.profiles.models import Profile
 
 
-@csrf_exempt
 def bio_view(request):
     if request.method == "POST":
         data = request.POST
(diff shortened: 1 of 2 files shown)


JRSiwiecki (approved):
Looks good to me!

mesh/profiles/views.py
```diff
@@ -1,6 +1,10 @@
+from django.core.exceptions import ObjectDoesNotExist
 from django.views.decorators.csrf import csrf_exempt
 from django.http import JsonResponse
 
+from mesh.profiles.models import Profile
+
+
 @csrf_exempt
```
owwix, line 8: We shouldn't be using @csrf_exempt tags in our code. It's a security risk if it gets into production. 

I should have disabled the CSRF middleware in the backend while we work on a proper CSRF implementation, eliminating the need for this tag. Have you tried not using it?

owwix (approved):
looks good for the most part, left a comment regarding one issue

Issue #25, opened 2023-03-30
A3.4 HTTP Get Call Handling


misslame:
﻿**As a User (Mentor, Mentee, or Educator), I want to be able to change my name as well as display my pronouns & nickname.** 

Details: Ensure that the back end can handle HTTP GET calls to allow for the fetching of the user's name, nickname, and pronouns. 

Technical Details: 
- Back-end
- MySQL
- HTTP GET
- Django

PR #271, opened 2023-08-01
Leaf/a3.4 http get call handling

shotoh:
issue #25

### DIFF
--- a/mesh/profiles/urls.py
+++ b/mesh/profiles/urls.py
@@ -4,8 +4,8 @@ from . import views as profile_views
 
 urlpatterns = [
     path("biography", profile_views.bio_view),
-    path("profile-picture", profile_views.ProfilePicturesView.as_view(), name="profile-picture"),
-    path("user-name", profile_views.UserNamesView.as_view(), name="user-name"),
-    path("preferred-name", profile_views.PreferredNamesView.as_view(), name="preferred-name"),
-    path("preferred-pronouns", profile_views.PreferredPronounsView.as_view(), name="preferred-pronouns"),
+    path("profile-picture/<int:account_id>", profile_views.ProfilePicturesView.as_view(), name="profile-picture"),
+    path("user-name/<int:account_id>", profile_views.UserNamesView.as_view(), name="user-name"),
+    path("preferred-name/<int:account_id>", profile_views.PreferredNamesView.as_view(), name="preferred-name"),
+    path("preferred-pronouns/<int:account_id>", profile_views.PreferredPronounsView.as_view(), name="preferred-pronouns"),
 ]
--- a/mesh/profiles/urls.py
+++ b/mesh/profiles/urls.py
@@ -4,8 +4,8 @@ from . import views as profile_views
 
 urlpatterns = [
     path("biography", profile_views.bio_view),
-    path("profilePicture", profile_views.ProfilePicturesView.as_view(), name="profile_picture"),
-    path("userName", profile_views.UserNamesView.as_view(), name="user_name"),
-    path("preferredName", profile_views.PreferredNamesView.as_view(), name="preferred_name"),
-    path("preferredPronouns", profile_views.PreferredPronounsView.as_view(), name="preferred_pronouns"),
+    path("profile-picture", profile_views.ProfilePicturesView.as_view(), name="profile-picture"),
+    path("user-name", profile_views.UserNamesView.as_view(), name="user-name"),
+    path("preferred-name", profile_views.PreferredNamesView.as_view(), name="preferred-name"),
+    path("preferred-pronouns", profile_views.PreferredPronounsView.as_view(), name="preferred-pronouns"),
 ]
--- a/mesh/profiles/urls.py
+++ b/mesh/profiles/urls.py
@@ -1,9 +1,11 @@
 from django.urls import path
+
 from . import views as profile_views
-from django.http import HttpRequest
 
 urlpatterns = [
-    
-    path('biography', profile_views.bio_view),
-    path("profilePicture", profile_views.profile_picture, name="profile_picture")
+    path("biography", profile_views.bio_view),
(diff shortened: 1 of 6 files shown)


mesh/profiles/views.py
```diff
@@ -21,26 +21,81 @@ def bio_view(request):
...
-            return JsonResponse(response)
-        else:
-            try:
-                profile = Profile.objects.get(accountID=int(data.get("accountID")))
-                response.update({
-                    "data": {
-                        "get": {
-                            "profilePicture": profile.image.url
-                        }
+        profile = get_profile(request.GET, response)
```
misslame, line 27: Setting up a response as a success before actually determining if its successful is bad practice.

mesh/profiles/views.py
```diff
@@ -21,26 +21,81 @@ def bio_view(request):
...
+        if profile is not None:
+            response.update({
+                "data": {
+                    "get": {
+                        "profilePicture": profile.image.url
+                    }
+                }
+            })
```
misslame, line 35: There are no error handlings. What if the profile is none? Right now it will return a success.

mesh/profiles/views.py
```diff
@@ -21,26 +21,81 @@ def bio_view(request):
...
+        return JsonResponse(response)
+
+
+def user_name(request):
+    if request.method == "GET":
```
misslame, line 40: Same as above comments. Error handling and communicating

mesh/profiles/views.py
```diff
@@ -21,26 +21,81 @@ def bio_view(request):
...
+                        "preferredPronouns": profile.preferredPronouns
+                    }
+                }
+            })
+        return JsonResponse(response)
+
+
+def get_profile(data, response):
+    if data.get("accountID") is None:
+        response.update({"status": "error"})
```
misslame, line 92: We should utilize exception handling here as previously done. Exception library json response formatting I believe was already done, it should be used.

misslame (changes requested):
I see some improvements that can be done.

misslame:
It still looks like you're doing repetitive error handling. 

Is a format like the following possible
try: 
   if valid get, 
       get the thing return response
  else 
       throw exception
except:

celinem13:
It's possible to handle most if not all error cases if you use a try catch. i believe you can do multiple tries and a single catch unless you want to specify whats the error.

misslame (approved):
A lot better! Woot

mesh/profiles/views.py
```diff
@@ -21,26 +22,43 @@ def bio_view(request):
...
+    if request.method == "GET":
+        return get_data(request, "preferredPronouns", lambda profile: profile.preferredPronouns)
+
+
+def get_data(request, name, mapper):
+    try:
+        data = validate_required_fields(request.GET, ["accountID"])
+        profile = Profile.objects.get(accountID=int(data["accountID"]))
+        return JsonResponse({
+            "status": "success",
```
misslame, line 48: Can you use http codes?

shotoh: done

JRSiwiecki (approved):
👍🏼

mesh/profiles/views.py
```diff
@@ -19,28 +21,45 @@ def bio_view(request):
...
+                "get": {
+                    name: mapper(profile)
+                }
+            }
+        }, status=200)
+    except ObjectDoesNotExist:
+        return JsonResponse({
+            "status": "error",
+            "message": "An account does not exist with this account ID."
+        }, status=400)
```
JRSiwiecki, line 60: Change this to a `404 Not Found` status probably.

mesh/profiles/urls.py
```diff
@@ -1,9 +1,11 @@
...
 
 urlpatterns = [
-    
-    path('biography', profile_views.bio_view),
-    path("profilePicture", profile_views.profile_picture, name="profile_picture")
+    path("biography", profile_views.bio_view),
+    path("profilePicture", profile_views.ProfilePicturesView.as_view(), name="profile_picture"),
+    path("userName", profile_views.UserNamesView.as_view(), name="user_name"),
+    path("preferredName", profile_views.PreferredNamesView.as_view(), name="preferred_name"),
+    path("preferredPronouns", profile_views.PreferredPronounsView.as_view(), name="preferred_pronouns"),
```
JRSiwiecki, line 10: This is purely a subjective ask, but can we try to use either kebab-case or snake_case in urls?

JRSiwiecki (changes requested):
I made some small comments, ping me on discord after you take a look at them so I can approve your PR.

Good work though! Great idea with the get_data method 👍🏼

mesh/utils/validate_data.py
```diff
@@ -34,8 +35,12 @@ def validate_json_and_required_fields(request_body, required_fields):
         data = json.loads(request_body)
     except json.JSONDecodeError:
         raise InvalidJsonFormat()
-        
+
+    return validate_required_fields(data, required_fields)
+
+
+def validate_required_fields(data, required_fields):
```
misslame, line 42: Just a question but why wouldn't you want to validate the json along with the fields? I see you use this method instead of using the json one...

misslame (approved):
Approving again with new changes. Just one question that doesn't seem incredibly important

JRSiwiecki (approved):

mesh/profiles/views.py
```diff
@@ -19,28 +19,39 @@ def bio_view(request):
...
+
+
+class PreferredPronounsView(View):
+    def get(self, request, account_id, *args, **kwargs):
+        return get_data(account_id, "preferredPronouns", lambda profile: profile.preferredPronouns)
+
+
+def get_data(account_id, name, mapper):
+    try:
+        profile = Profile.objects.get(accountID=int(account_id))
```
misslame, line 44: Here is where you'd call the validate json, and youd pass in the request body im pretty sure

Issue #139, opened 2023-05-20
First Sprint Front End Head Aches


misslame:
Combine the work for first sprint. Please work with the following task assignees to get their work in together:

- [x] #138 
- [x] #136 
- [x] #135 
- [x] #134 
- [x] #114 
- [x] #76 
- [x] #64 
- [x] #2

PR #272, opened 2023-08-06
Construct pages with routing

Darren52BT:
This addresses #139.

# What I've Done
I have basically created a shell for the website.  I used React Router to create routes for each page. Upon loading, you should see a temporary list of links that I put down to make it easier to navigate around these routes. You should also be able to click on links on various pages to go to their respective pages. I've only done so for links that we have page components for, so the other links do nothing.

# Structure 
Currently, there are three sets of routes that are stacked on top of each other. The first defines routes for which header to display, the second for which main page component to display ~~, and the third for which footer to display.~~  (Edit: I've changed it to always display the footer with links, and to show the messages appbar fixed at the bottom if we're at a logged out page). I renamed and moved the various header and footer components to separate folders to make it more organized. I also created another header component for the logged out pages aside from the logged out home page.

All route paths are being defined within an object called paths that is being used globally to reference to a certain path. If we want to change the path for a certain page globally, all we need to do is edit this object.

I've edited the various components representing links to function as the Link component from React Router.

# Notes
-as this is a shell, this is just showing which components to display for the current route. We will need to rework it further later to pass data between pages.
-I think that there is supposed to be a separate page for the forgot password page. I have defined a page for it but it does not completely fill up the screen so it looks pretty awkward. Not sure if I should take care of it here or if that should be done separately.

# What You Should see
## Default Page (containing temp link list)
image

## Ex: Clicking on Logged Out Home Page Link:
image

image

## Ex: Clicking on Swiping Link in Logged In Header:
image
image


meshapp/src/Routing/FooterRoutes.tsx
```diff
@@ -0,0 +1,31 @@
...
+  paths.sign_up,
+];
+
+//contains routes for the footer
+export default function FooterRoutes() {
+  return (
+    <Routes>
+      {LOGGED_IN_PATHS.map((path) => (
+        <Route path={path} element={<LoggedInFooter />} />
+      ))}
```
misslame, line 25: We have different footers based on login?

meshapp/src/Routing/HeaderRoutes.tsx
```diff
@@ -0,0 +1,30 @@
...
+const LOGGED_OUT_PATHS = [
+  //   paths.logged_out_home,
+  paths.login_page,
+  paths.forgot_password,
+  paths.sign_up,
+];
+
+export default function HeaderRoutes() {
+  return (
+    <Routes>
```
misslame, line 21: Something seems fishy but idk how to fix, why are header and footer files essentially the same?

meshapp/src/home/login.tsx
```diff
@@ -73,20 +78,21 @@ const Login = () => {
             bgcolor: "cardBackground.main",
             color: "text.main",
             borderRadius: 5,
-          }}>
+          }}
+        >
```
misslame, line 82: Nit picky but why new line

misslame (approved):
Approving but I have comments

meshapp/src/Routing/HeaderRoutes.tsx
Darren52BT: Well they are pretty much doing the same thing, which is showing a specific version of the header/footer for logged in/logged out pages, but the header for the logged out home page is different from the other logged out pages, so there's a difference in routes.

meshapp/src/Routing/FooterRoutes.tsx
Darren52BT: The footer with the message icon is different from the one with all the links right? That's what I'm seeing on the figma

meshapp/src/Routing/MainPageRoutes.tsx
```diff
@@ -0,0 +1,58 @@
+import LoggedOutPage from "../home/logged-out/home";
+import SignUp from "../components/SignUp/SignUp";
+import Login from "../home/login";
+import ForgotPassword from "../components/password-forms/forgot-password-form";
+import LoggedInHome from "../home/logged-in/LoggedInHome";
+import ProfilePage from "../profile/profile-page";
+import {
+  exampleProfile,
+  exampleProfile2,
```
JRSiwiecki, line 9: Is exampleProfile2 being used anywhere else here?

JRSiwiecki (commented):
Everything seems mostly fine to me, I just have a question: when I was viewing the site on your branch, it seemed that for some pages like the forgot_password page or signup page, the home link doesn't lead anywhere, is that correct? I can understand if that's just confusion over whether to send people to the logged in home or logged out home, but I think as a nice default it would just send people to logged out home (or even back to the main routing page).

Other than those questions, I really don't have much to say other than good work! Please re-ping me on discord so I can send approval! 🥇

meshapp/src/Routing/FooterRoutes.tsx
misslame: The message icon shouldn't be a part of the footer it should be by itself

meshapp/src/Routing/HeaderRoutes.tsx
misslame: Hmmmm

misslame: Ill come back to this

meshapp/src/Routing/FooterRoutes.tsx
Darren52BT: It's shown consistently along all the logged in pages at the bottom though so shouldnt it be a footer? should the footer with the links be included in the logged in pages as well?

misslame: If only one thing is different, the whole footer shouldn't be rendered as a different footer

misslame: Its showed down there. It should be like the facebook message bar

Darren52BT: So is it meant to be something that is fixed at the bottom and appears over other items similarly to the 2FA banner? In what other ways should I be displaying this for the logged in pages?

PR #276, opened 2023-08-09
Build and Run commands

Leaversa:
Adds new commands to build and run the project from the root directory:

`npm run init`: Django migrations and installs npm packages for frontend
`npm run build`: builds the frontend
`npm run start`: starts the Django server

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,12 @@
+{
+  "name": "letsmesh",
+  "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


owwix (approved):
Looks good to me

peppacaiou (approved):
Code looks logical and sound, If this has been tested before, this loosks great. If it hasn't been tested before, please do so :)

Issue #81, opened 2023-04-25
B5.3 HTTP Get Call Handling 


misslame:
Story: **As an account owner, I want to be able to edit account settings and information such as email, password, enable matching, etc.**

Details: When the client fetches information for populating the settings UI, handle the GET request. 

Technical Details:
- Back-end
- MySQL
- Django
- Python
- HTTP REST Get

PR #277, opened 2023-08-10
B5.3 HTTP Get Call Handling

tdang2180:
#81 
- Implemented GET call handling for fetching account settings
- Added tests for account settings

Not sure if this satisfies the task, but from my understanding of the task's description, I think this is what I had to do.

### DIFF
--- a/mesh/accountSettings/views.py
+++ b/mesh/accountSettings/views.py
@@ -1,8 +1,8 @@
 from django.core.exceptions import ObjectDoesNotExist
 from django.http import JsonResponse
-
+from django.views import View
+from django.core import serializers
 from mesh.accountSettings.models import Settings
-
 
 def display_theme(request):
     if request.method == "GET":
@@ -29,3 +29,25 @@ def display_theme(request):
                 response.update({"status": "error"})
                 response.update({"message": "An account does not exist with this account ID."})
                 return JsonResponse(response)
+
+class SettingsView(View):
+    def get(self, request, *args, **kwargs):
+        """
+        Handles GET requests when the client fetches for all accounts settings
+        """
+        settings = Settings.objects.all()
+        accounts_settings = serializers.serialize('json', settings)
+        return JsonResponse(accounts_settings, safe=False)
+
+class SettingsDetailView(View):
+    def get(self, request, account_id, *args, **kwargs):
+        """
+        Handles GET requests when the client fetches for a specific account settings
+        """
+        try: 
+            settings = Settings.objects.get(accountID= account_id)
+            settings_detail = serializers.serialize('json', [settings])
+            return JsonResponse(settings_detail, safe = False)
+        
+        except Settings.DoesNotExist:
+            return JsonResponse({'error': 'Settings for this account does not exist'}, status=404)
--- a/mesh/accountSettings/views.py
+++ b/mesh/accountSettings/views.py
@@ -37,7 +37,7 @@ class SettingsView(View):
         """
         settings = Settings.objects.all()
         accounts_settings = serializers.serialize('json', settings)
-        return JsonResponse(accounts_settings, safe=False)
+        return JsonResponse(accounts_settings, safe=False, status=200)
 
 class SettingsDetailView(View):
     def get(self, request, account_id, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


BradJKim:
Hihi, I'm super new to code reviewing and I'm just trying to take part in any shape or form. I took a look and everything seemed to be all set. I just had a question about how the testing worked and was wondering if you could explain the code cause I personally I'm not familiar with that. Other than that, it all seems really nice.

tdang2180:
Hi @BradJKim!! For this task, I used Django's class-based views API to handle the GET request whenever the client fetches information for the settings. Instead of creating a JSON array to store the response, I used Django's serialization to 'translate' the settings model to JSON format. I also added URLs to give the browser access to both of the views. For the testing, I just created test cases to check if the status code of the GET response matches the successful status code. Hope this explanation makes sense because I am also new to this.

mesh/accountSettings/views.py
```diff
@@ -29,3 +29,25 @@ def display_theme(request):
...
+
+class SettingsDetailView(View):
+    def get(self, request, account_id, *args, **kwargs):
+        """
+        Handles GET requests when the client fetches for a specific account settings
+        """
+        try: 
+            settings = Settings.objects.get(accountID= account_id)
+            settings_detail = serializers.serialize('json', [settings])
+            return JsonResponse(settings_detail, safe = False)
```
JRSiwiecki, line 50: Make sure to include HTTP status codes in this return and the one to GET all AccountSettings.

mesh/accountSettings/views.py
```diff
@@ -29,3 +29,25 @@ def display_theme(request):
...
                 response.update({"message": "An account does not exist with this account ID."})
                 return JsonResponse(response)
+
+class SettingsView(View):
+    def get(self, request, *args, **kwargs):
+        """
+        Handles GET requests when the client fetches for all accounts settings
+        """
+        settings = Settings.objects.all()
+        accounts_settings = serializers.serialize('json', settings)
```
JRSiwiecki, line 39: Maybe it's just me, but I somewhat dislike how Django's built-in serializer serializes the data, as even on my task which is very similar to this, using the built-in serializer returns just a not so pretty looking string.

For example, when I test this end-point, the output I get looks like this:
```'[{"model": "accountSettings.settings", "pk": 1, "fields": {"isVerified": true, "verificationToken": "1", "hasContentFilterEnabled": true, "displayTheme": "1", "is2FAEnabled": true}}]'```

but if I change the serializer code to be more like this (similar to how I went about this "issue" in my task):

```
    accounts_settings = [
            {
                "accountID": setting.accountID.accountID,
                "isVerified": setting.isVerified,
                "VerificationToken": setting.verificationToken,
                "HasContentFilterEnabled": setting.hasContentFilterEnabled,
                "DisplayTheme": setting.displayTheme,
                "Is2FAEnabled": setting.is2FAEnabled
            }
            for setting in settings
        ]
```

it will output like this:
```
[{"accountID": 1, "isVerified": true, "VerificationToken": "1", "HasContentFilterEnabled": true, "DisplayTheme": "1", "Is2FAEnabled": true}]
``` 

which is much prettier to look at if you ask me.

I have some examples of what it looks like in a thread about this specific B5.3 issue if you want to see screenshots of what it actually looks like, but currently I'm waiting for Bob's response or even Chris' response if he reads this before Bob.

For the time being I think the way you have it is fine and can stay that way, but I'm going to hold off on approving until I can get Bob or Chris' opinion.

Other than that though, good work 👍🏼

JRSiwiecki (commented):
This is very well done other than a few small things like the HTTP status codes, but I'm going to hold off on approval until I can get an opinion from Bob or Chris.

mesh/accountSettings/views.py
JRSiwiecki: To clarify I don't necessarily enjoy this solution as much cause it is somewhat manual and to do this for every model would be painstaking, but I'm open to other ideas 👍🏼

JRSiwiecki: Update from Chris is that it is probably fine to use Django Serializer, so I think what you have is good!

JRSiwiecki (approved):
Good work 🥇

mesh/accountSettings/views.py
tdang2180: 👍

shotoh (approved):
Looks good to me

Issue #116, opened 2023-04-25
B4.3	Error Handling on Login page


GuyWhoCode:
Story: **As a User (Mentor, Mentee, or Educator), I want to be able to log in with my account so I can start using the service..**

Details: UI in the login page to inform the user of various errors like an incorrect password or email or too many login attempts. 

Technical Details:
- React
- HTTP REST Post response handling
- UX
- UI
- Axios

PR #278, opened 2023-08-11
Leaf/b4.3 error handling on login page

owwix:
Completes issue B4.3 Error Handling on Login page #116
Screen Shot 2023-08-10 at 5 52 39 PM


JRSiwiecki:
Make sure you are merging into `dev_new` not `main`.

peppacaiou (approved):
The code looks great, not too intuitive for someone like me, but that's because I've never had a chance to work with Djoser yet. A few more comments specifying niche code wouldn't hurt, but logic is great and straightforward

mesh/settings.py
```diff
@@ -42,6 +42,8 @@
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
+    'rest_framework',
+    'djoser',
```
JRSiwiecki, line 46: These should be added to the pipfile if it's required for the project, right?

meshapp/src/components/login-form.tsx
```diff
@@ -111,8 +133,8 @@ const LoginScreen = (props: ComponentProps) => {
 }
 
 const ForgotPasswordScreen = (props: ComponentProps) => {
-
-  const forgotPasswordEndpoint: string = '/user/reset';
+  
+  const forgotPasswordEndpoint: string = 'http://localhost:8000/accounts/auth/users/reset_password/';
```
JRSiwiecki, line 137: Small nitpick, but might make more sense to just have a string that has the URL domain, and then just have the endpoint appended to that URL so that we can just change the URL domain string in the future.

mesh/settings.py
```diff
@@ -153,11 +155,42 @@
...
-]
\ No newline at end of file
+# Djoser (User Management, Authentication, etc)
+# https://djoser.readthedocs.io/en/latest/introduction.html
+
+DJOSER = {
+    'LOGIN_FIELD':'email',
+    'USER_CREATE_PASSWORD_RETYPE': True,
+    'PASSWORD_CHANGED_EMAIL_CONFIRMATION': True,
+    'SEND_CONFIRMATION_EMAIL': True,
```
JRSiwiecki, line 176: I'm just curious about this is I have never used Djoser or seen much about it, but is it correct to believe that Djoser could handle email confirmation for accounts?

JRSiwiecki (commented):
Everything seems good based on my tiny bit of research of Djoser, but please review my feedback and then re-ping me on Discord for approval.

mesh/settings.py
owwix: Yes, most definitely

mesh/settings.py
owwix: Correct, just fixed that. Thank you!

meshapp/src/components/login-form.tsx
owwix: There actually isn't a need for the URL domain at all since front and backend are ran on the same port. For example, look at my LoginScreen component.

I plan on fixing this in my next PR. Since it involves making the forgot password form fully functional.

JRSiwiecki (approved):
👍🏼

PR #279, opened 2023-08-11
Finished mysql readme documentation.

peppacaiou:
Wrote the docs for the MySQL setup.
Currently, the readMe has been tested and works for windows, but am not sure about mac and linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ MYSQL SETUP (windows, if you have a mac, consider switching to a windows)
 4. Spam next until you finish and then you’re done
 Make sure you write the username and password in your.env file. With that being said,that’s it for your mysql setup, next thing we do is django.
 
-1. Open a terminal and make sure mysql is installed by typing in `mysql --v`
+1. Open a terminal and make sure mysql is installed by typing in `mysql -V`
     a. If the command doesn’t work it may be because mysql isn’t added to your path. The default location for mysql server should be: 
         C:\Program Files\MySQL\MySQL Server 8.0\bin\mysql.exe
             i. Possible solution: https://stackoverflow.com/questions/5920136/mysql-is-not-recognised-as-an-internal-or-external-command-operable-program-or-b
@@ -92,8 +92,8 @@ Make sure you write the username and password in your.env file. With that being 
         i. Do note that root is the default username so use whatever username you used
     b. You should be prompted into they mysql shell, and then create a database by using `CREATE DATABASE mesh;`
     c. Confirm your database is created with `SHOW DATABASES;`
-3. Open you pip environment is opened (with `pipenv shell`)
-4. Type in `python manage.py makemigrations` 
+3. make sure your pip environment is opened (with `pipenv shell`)
+4. Type in `python manage.py makemigrations`
     a. You will most likely be given warnings. Warnings are fine errors aren’t. Google your errors.
 5. Type in `python manage.py migrate`
     a. Similar to the previous step warnings are fine, errors aren’t.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -75,6 +75,29 @@ Check npm version with `npm -v` -> This should show 9.6.7 in the console
...
+    a. If the command doesn’t work it may be because mysql isn’t added to your path. The default location for mysql should be: 
+        C:\Program Files\MySQL\MySQL Server 8.0\bin\mysql.exe
+            i. Possible solution: https://stackoverflow.com/questions/5920136/mysql-is-not-recognised-as-an-internal-or-external-command-operable-program-or-b
+2. Once you have mysql downloaded, create a new database inside your mysql.
+    a. Login to mysql by using  `mysql --u root --p` and then type in the password you saved previously when prompted
+        i. Do note that root is the default username so use whatever username you used
+    b. You should be prompted into they mysql shell, and then create a database by using `CREATE DATABASE mesh;`
+    c. Confirm your database is created with `SHOW DATABASES;`
+3. Open you pip environment is opened (with `pipenv shell`)
+4. Type in python `manage.py makemigrations`
```
JRSiwiecki, line 96: Should we be having everyone do makemigrations? My understanding is that you do makemigrations to create the actual migrations files in the migrations folder for each model, then run migrate to apply those migrations. Having each person make migration files should not be necessary, as you would only need to makemigrations if someone was making changes to existing models.

People who clone the repo should be able to just do migrate immediately assuming they created their database, as it would just use the existing migration files.

JRSiwiecki (commented):
I wasn't requested to review this but I have some thoughts 💯

README.md
```diff
@@ -75,6 +75,29 @@ Check npm version with `npm -v` -> This should show 9.6.7 in the console
...
+1. Open a terminal and make sure mysql is installed by typing in `mysql --v`
+    a. If the command doesn’t work it may be because mysql isn’t added to your path. The default location for mysql server should be: 
+        C:\Program Files\MySQL\MySQL Server 8.0\bin\mysql.exe
+            i. Possible solution: https://stackoverflow.com/questions/5920136/mysql-is-not-recognised-as-an-internal-or-external-command-operable-program-or-b
+2. Once you have mysql downloaded, create a new database inside your mysql.
+    a. Login to mysql by using  `mysql --u root --p` and then type in the password you saved previously when prompted
+        i. Do note that root is the default username so use whatever username you used
+    b. You should be prompted into they mysql shell, and then create a database by using `CREATE DATABASE mesh;`
+    c. Confirm your database is created with `SHOW DATABASES;`
+3. Open you pip environment is opened (with `pipenv shell`)
```
JRSiwiecki, line 95: Also this sentence should probably be more clear about what it is saying 💯

README.md
peppacaiou: From my experiences it doesn't seem to hurt when people do "make migrations", it does hurt if migrations aren't made. Which is why i just told them to make migrations before migrate

README.md
```diff
@@ -75,6 +75,29 @@ Check npm version with `npm -v` -> This should show 9.6.7 in the console
 6. Go to `Site/meshapp/` and run an `npm install`
 7. Build the frontend -> in the same directory as step 7 run `npm run build`
 8. Return to the root directory and start the backend server / run the application -> `python manage.py runserver`
+MYSQL SETUP (windows, if you have a mac, consider switching to a windows)
+1. Download Mysql over here https://dev.mysql.com/downloads/installer/ ( i downloaded the fatter file)
```
BradJKim, line 79: Being specific might help lol, not just fatter file

BradJKim, line 79: I think begin more specific will help lol, instead of just fatter file

BradJKim (changes requested):
This seems good, some things can be a bit more specific, but the rest is fine.

There also seems to be like the dot_env error happening where we had to remove the two lines, I don't know if you remember. I'm not sure if it's like your job to address it, but I think it be nice and would prevent new people from being stuck.

JRSiwiecki (approved):
RAHHHHHHH

BradJKim (approved):

BradJKim (approved):

PR #284, opened 2023-08-14
Implement Profile Picture: HTTP Post Call Handling

### LINKED ISSUE #7: A1.7 Profile Picture: HTTP Post Call Handling
﻿**Story:** ﻿As a User (Mentor, Mentee, or Educator), I want to be able to upload a professional profile picture of myself or something representing me.
﻿
﻿﻿**Details:**
﻿Allow for the retrieval of an image to be stored and for the previous image to be deleted within the database.
﻿
﻿Note: ﻿Allow for the removal of the image if no image is provided and the user specified removing the image, replace with default image.

﻿**Technical Information:**
- Django
- Handle Post Call
- Error Handling as a response is expected for improper Post Calls
- Back-end

JRSiwiecki:
Resolves #7 

**FEATURES:**

- Implemented endpoints for managing user profile pictures.

**ProfilePicturesView:**

- Added POST request endpoint (`POST /profile-pictures/`) to upload a profile picture and link it with the user's profile.
- The POST request requires the following fields: `accountID` and `profilePicture`.
- Checks if the user already has a profile picture. If yes, returns a `409 Conflict` response with an error message.
- Otherwise, encodes and uploads the image to a Backblaze B2 Bucket, retrieves the image link, and saves it with the user's profile.
- Returns a JSON response with the `profileID` and `profilePicture` links.

- Added PATCH request endpoint (`PATCH /profile-pictures/`) to update a user's profile picture.
- The PATCH request requires the following fields: `accountID` and `profilePicture`.
- Checks if the user has an existing profile picture. If not, returns a `409 Conflict` response with an error message.
- Otherwise, encodes and uploads the new image to a Backblaze B2 Bucket, retrieves the new image link, and updates it in the user's profile.
- Returns a JSON response with the `profileID` and updated `profilePicture` link.

- Added DELETE request endpoint (`DELETE /profile-pictures/`) to delete a user's profile picture.
- The DELETE request requires the field: `accountID`.
- Checks if the user has an existing profile picture. If not, returns a `404 Not Found` response with an error message.
- Otherwise, deletes the profile picture link from the user's profile.
- Returns a `204 No Content` response.

**ENDPOINTS:**

**POST /profile-pictures/**
- Uploads a profile picture and links it with the user's profile. Requires fields: `accountID` and `profilePicture`.

**PATCH /profile-pictures/**
- Updates a user's profile picture. Requires fields: `accountID` and `profilePicture`.

**DELETE /profile-pictures/**
- Deletes a user's profile picture. Requires field: `accountID`.

Note: Most migrations files can be ignored, the only change is the date they were created, except for the migrations for profile, which needed to be re-made to update existing tables.


mesh/profiles/views.py
```diff
@@ -44,3 +60,169 @@ def profile_picture(request):
...
+        except HTTPError:
+            return JsonResponse({"error": "Uploaded file is not an image."}, status = 400)
+
+        except ValueError:
+            return JsonResponse({"error": "accountID or profilePicture field is empty."}, status = 400)
+
+    def patch(self, request, *args, **kwargs):
+
+        try:
+            data = MultiPartParser(request.META, request, request.upload_handlers).parse()
```
shotoh, line 136: Can you explain what this is?
Also, for POST you used request.POST and request.FILES while in PATCH, you used whatever request.META and MultiPartParser
Then you used validate_json_and_required_fields and request.body for the DELETE request

What's the difference between the three?

shotoh (changes requested):

mesh/profiles/views.py
```diff
@@ -44,3 +60,169 @@ def profile_picture(request):
...
+            account_id = data["accountID"]
+            
+            profile = Profile.objects.get(accountID=account_id)
+            
+            # Check if user already has a profilePicture
+            if profile.profilePicture is not None:
+                return JsonResponse({"error": "Profile already has a profilePicture. " +
+                                     "Use PATCH to update the profilePicture or " + 
+                                     "DELETE to delete the profilePicture."}, 
+                                    status = 409)
```
shotoh, line 99: Shouldn't the status code be 405?

mesh/profiles/views.py
```diff
@@ -44,3 +60,169 @@ def profile_picture(request):
...
+        try:
+            data = MultiPartParser(request.META, request, request.upload_handlers).parse()
+            
+            account_id = data[0]["accountID"]
+
+            profile = Profile.objects.get(accountID=account_id)
+
+            if profile.profilePicture is None:
+                return JsonResponse({"error": "Profile does not have an existing profilePicture." +
+                                     "Use a POST request to upload a profilePicture."}, status = 409)
```
shotoh, line 144: Again shouldn't the status code be 405?

mesh/profiles/views.py
```diff
@@ -44,3 +60,169 @@ def profile_picture(request):
...
+            image_link = imgur_response["link"]
+
+            # Save photo URL with user
+            profile.profilePicture = image_link
+            profile.save()
+
+            request_response = json.dumps({"profileID": profile.accountID.accountID ,
+                                 "profilePicture": image_link})
+            
+            return JsonResponse(request_response, status = 201, safe = False)
```
shotoh, line 166: Shouldn't the status code be 200 since it's updating instead of creating a new resource?

shotoh (changes requested):
My previous review didn't include these comments below for some reason so I just made another review

mesh/profiles/views.py
JRSiwiecki: My thought is that it should be 409 rather than 405, as "The HTTP 409 status code (Conflict) indicates that the request could not be processed because of conflict in the request, such as the requested resource is not in the expected state, or the result of processing the request would create a conflict within the resource."

405 would mean that we recognize the request method, but it isn't allowed, which isn't the case as we do what to allow users to POST or PATCH.

mesh/profiles/views.py
JRSiwiecki: Same as above.

mesh/profiles/models.py
```diff
@@ -1,15 +1,13 @@
...
-# Create your models here.
-
 class Profile(models.Model):
     accountID = models.ForeignKey(Account, primary_key=True, on_delete=models.CASCADE)
     userName = models.CharField(max_length= 255)
     preferredName = models.CharField(max_length= 255, null= True)
     preferredPronouns = models.CharField(max_length = 40)
     biography = models.TextField()
-    image = models.ImageField(upload_to = "image")
+    profilePicture = models.URLField(max_length = 200, null = True, blank = True)
```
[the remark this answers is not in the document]
JRSiwiecki: 👍🏼

mesh/profiles/views.py
JRSiwiecki: Yea that sounds right, I'll fix this!

mesh/profiles/views.py
JRSiwiecki: So for PATCH requests, a file will be uploaded, but we can't use request.POST to get the form data, because we aren't receiving a POST request. So a work-around is to use the MutliPartParser that Django provides, as we still need an easy way to grab the files uploaded.

The POST request and PATCH requests both should use forms when sent, because you can't send files over JSON, right? So we need to use request.POST/request.FILES/MultiPartParser to actually access the uploaded files in their respective request methods. 

The DELETE request DOES use JSON, so we can use the validate_json_and_required_fields method, but the other two do NOT use JSON, so we don't have access to that method.

mesh/profiles/models.py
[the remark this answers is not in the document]
JRSiwiecki: I actually changed this for you since I had to fix merge conflicts 👍🏼

tdang2180:
Hi @JRSiwiecki, I wasn't requested to review this PR, but I'm doing a similar task and I am confused about how the profile is set up. I looked at your code and noticed that you have code to check whether a profile exists in your POST method, but how would a profile exist if there is no method to create a profile in the first place?  

I feel like there should be a task for implementing a POST method that creates a profile with all the fields and tasks like these to patch/change the individual fields. I don't know if this makes sense but I am confused because I don't see any task (HTTP post call handling) for creating a profile.

shotoh (approved):
Everything else looks fine to me

mesh/profiles/urls.py
```diff
@@ -2,12 +2,25 @@
...
     # GET    /api/biography/:account_id
     # POST   /api/biography/:account_id
     path('biography/<int:account_id>/', profile_views.BiographyView.as_view(), name='biography'),
-    path("profilePicture", profile_views.profile_picture, name="profile_picture"),
-    path("profile-picture/<int:account_id>", profile_views.ProfilePicturesView.as_view(), name="profile-picture"),
+    
+    # GET /profiles/profile-picture/:account_id
+    path("profile-picture/<int:account_id>", profile_views.ProfilePictureDetailsView.as_view(), name="profile-picture-details"),
+    
+    # GET /profiles/profile-picture/:account_id
```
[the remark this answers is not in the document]
JRSiwiecki: oh true i literally copy pasted them 💀

Darren52BT (approved):
lgtm

PR #285, opened 2023-08-15
A2.4 HTTP Get Call Handling

BradJKim:
- Implemented Get handling for fetching profile biography
- Added test for profile biography

This is my first pull request so there are probably a few errors in the code. The main files I worked on was urls.py and view.py in profiles, and profiles_test.py. Would love to get any feedback I can.

I'm also not sure if we need to test for different errors as well when testing, I only created one that would just retrieve the biography under expected conditions.

### DIFF
--- a/mesh/profiles/urls.py
+++ b/mesh/profiles/urls.py
@@ -1,9 +1,12 @@
-from django.urls import path
+from django.urls import path, include
 from . import views as profile_views
 from django.http import HttpRequest
+from .views import *
 
 urlpatterns = [
-    
-    path('biography', profile_views.bio_view),
+    # GET    /api/biography/:account_id
+    # POST   /api/biography/:account_id
+    path('biography/<int:account_id>/', profile_views.BiographyView.as_view(), name='biography'),
+
     path("profilePicture", profile_views.profile_picture, name="profile_picture")
 ]
--- a/mesh/profiles/urls.py
+++ b/mesh/profiles/urls.py
@@ -1,7 +1,6 @@
 from django.urls import path, include
 from . import views as profile_views
 from django.http import HttpRequest
-from .views import *
 
 urlpatterns = [
     # GET    /api/biography/:account_id
--- a/mesh/profiles/views.py
+++ b/mesh/profiles/views.py
@@ -1,22 +1,49 @@
 from django.core.exceptions import ObjectDoesNotExist
-from django.views.decorators.csrf import csrf_exempt
 from django.http import JsonResponse
+from django.core import serializers
+from mesh.accounts.models import Account
 
 from mesh.profiles.models import Profile
 
+from django.views import View
 
-def bio_view(request):
-    if request.method == "POST":
-        data = request.POST
-        if 'accountID' not in data:
-            return JsonResponse({'error': 'Invalid request. Missing accountID field.'}, status=401)
-        if 'biography' not in data:
-            return JsonResponse({'error': 'Invalid request. Missing biography field.'}, status=400)
+class BiographyView(View):
+    """
+    View for getting an biography by accountId or updating an biography by accountID.
+    """
+    def get(self, request, account_id, *args, **kwargs):
+        """
+        Handle GET requests.
+
+        Retrieves a biography of the listed profile in JSON format.
+
+        Returns a JSON response containing biography of specified profile through id.
+        """
+        try:     
+            if(account_id is None):
+                return JsonResponse({'error': 'Invalid input. Missing account ID'}, status=401)
+            
+            profile = Profile.objects.get(accountID=account_id)
+            profileBiography = profile.biography
+            return JsonResponse({'message': profileBiography}, safe=False, status=200)
+        except Account.DoesNotExist:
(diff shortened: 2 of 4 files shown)


mesh/profiles/views.py
```diff
@@ -1,22 +1,49 @@
...
+            if(account_id is None):
+                return JsonResponse({'error': 'Invalid input. Missing account ID'}, status=401)
+            
+            profile = Profile.objects.get(accountID=account_id)
+            profileBiography = profile.biography
+            return JsonResponse({'message': profileBiography}, safe=False, status=200)
+        except Account.DoesNotExist:
+            return JsonResponse({'error': 'Invalid request. Account does not exist'}, status=404)
+        except Profile.DoesNotExist:
+            return JsonResponse({'error': 'Invalid request. Profile does not exist'}, status=402)
```
JRSiwiecki, line 32: This status code should likely also be `status = 404` since 402 is `402 Payment Required`.

mesh/profiles/views.py
```diff
@@ -1,22 +1,49 @@
...
+        Retrieves a biography of the listed profile in JSON format.
+
+        Returns a JSON response containing biography of specified profile through id.
+        """
+        try:     
+            if(account_id is None):
+                return JsonResponse({'error': 'Invalid input. Missing account ID'}, status=401)
+            
+            profile = Profile.objects.get(accountID=account_id)
+            profileBiography = profile.biography
```
JRSiwiecki, line 27: Small nitpick, but python variable naming conventions should usually follow `snake_case` rather than `camelCase`.

mesh/profiles/urls.py
```diff
@@ -1,9 +1,12 @@
-from django.urls import path
+from django.urls import path, include
 from . import views as profile_views
 from django.http import HttpRequest
+from .views import *
```
JRSiwiecki, line 4: The `from .views import *` seems like it would be redundant when `from . import views as profile_views`

JRSiwiecki (changes requested):
Other than some small things that could be changed, this looks good!

Ping me on discord after you've taken a look at my feedback and I'll approve 👍🏼

mesh/profiles/views.py
```diff
@@ -1,22 +1,49 @@
...
+        """
+        Handle GET requests.
+
+        Retrieves a biography of the listed profile in JSON format.
+
+        Returns a JSON response containing biography of specified profile through id.
+        """
+        try:     
+            if(account_id is None):
+                return JsonResponse({'error': 'Invalid input. Missing account ID'}, status=401)
```
misslame, line 24: Throw an exception instead of doing error handling with a jspn response and *sometimes* using exceptions. 

Exceptions are for error handling.

mesh/profiles/views.py
```diff
@@ -1,22 +1,49 @@
...
+
+    def post(self, request, *args, **kwargs):
+        if request.method == "POST":
+            data = request.POST
+            if 'accountID' not in data:
+                return JsonResponse({'error': 'Invalid request. Missing accountID field.'}, status=401)
+            if 'biography' not in data:
+                return JsonResponse({'error': 'Invalid request. Missing biography field.'}, status=400)
+            else:
+                print(data.get('accountID') + " " + data.get('biography'))  # TODO: Save to database
```
misslame, line 43: Same situation here. I would also use the exceptions in the exceptions folder and make new ones if you need custom error handling.

mesh/profiles/urls.py
misslame: I generally dont suggest importing all. I suggest importing only what's necessary.

misslame (changes requested):

mesh/profiles/views.py
JRSiwiecki: I don't think Brad wrote this part, if I recall correctly this was covered in #206. It looks like he just moved the method written from that issue into the BiographyView class.

mesh/profiles/views.py
JRSiwiecki: Wouldn't it actually be better to use the `validate_json_and_required_fields` from the utils folder here and just catch the InvalidJsonFormat/MissingRequiredFields exceptions raised from that function, rather than raising an exception himself?

misslame: Yea it would @JRSiwiecki

mesh/profiles/views.py
misslame: Ohh

misslame: Why is it showing up as his change?

BradJKim: I moved it into the Biography view class. Previously, the code was just its own method, similarly to the get Profile picture method below in the same file. I thought it would make sense if it put the post method in with the get method I was doing (I think that makes sense?? dunno). 

Should I also change the post request to make it consistent with the get request I wrote?

mesh/profiles/views.py
JRSiwiecki: Actually I just realized this isn't even json it's from the URL so that method would not work here, my bad 👎🏼

JRSiwiecki (approved):
👍🏼

misslame (approved):

Issue #26, opened 2023-03-30
A3.5 HTTP Post Call Handling


misslame:
﻿**As a User (Mentor, Mentee, or Educator), I want to be able to change my name as well as display my pronouns & nickname.** 
﻿
﻿Details: Ensure that the back end can handle HTTP POST Calls to allow for the sending of names/nicknames/pronouns from the front end. 
﻿
﻿Technical Details:
- ﻿HTTP POST 
- Back-end
- MySQL
- Django

PR #290, opened 2023-08-20
A3.5 HTTP Post Call Handling

tdang2180:
#26 
- Implemented POST call handling and added tests for name/nickname/pronouns.


mesh/tests/profiles_tests.py
```diff
@@ -89,6 +89,21 @@ def test_no_account_preferred_pronouns(self):
...
+        user_name_data = {"userName": "kwame brown"}
+        response = self.client.post(f'/profiles/user-name/{self.test_profile.accountID}', user_name_data)
+        self.assertEquals(response.status_code, 200)
+
+    def test_post_preferred_name (self):
+        preferred_name_data = {"preferredName": "brown"}
+        response = self.client.post(f'/profiles/preferred-name/{self.test_profile.accountID}', preferred_name_data)
+        self.assertEquals(response.status_code, 200)
+
+    def test_post__preferred_pronouns (self):
```
shotoh, line 102: Pretty sure you didn't mean to put two underscores here

shotoh (approved):
Everything else looks fine to me

mesh/tests/profiles_tests.py
tdang2180: 👍

mesh/profiles/views.py
```diff
@@ -77,3 +85,31 @@ def get_data(account_id, name, mapper):
             "status": "error",
             "message": "An account does not exist with this account ID."
         }, status=404)
+
+def post_data(account_id, name, request):
+    """
+    Handles POST requests when the client sends names/nicknames/prounouns to the back end
+    """
+    try: 
+        account = Account.objects.get(accountID = account_id)
```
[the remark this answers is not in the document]
tdang2180: Good point, I removed it. 👍

mesh/profiles/views.py
```diff
@@ -1,9 +1,10 @@
 from django.core.exceptions import ObjectDoesNotExist
+from django.utils.datastructures import MultiValueDictKeyError
 from django.http import JsonResponse
 
 from mesh.accounts.models import Account
 from mesh.profiles.models import Profile
-
+from mesh.accounts.models import Account
```
JRSiwiecki, line 7: Probably remove one of these duplicate imports.

mesh/profiles/views.py
```diff
@@ -77,3 +85,27 @@ def get_data(account_id, name, mapper):
...
+
+        if (name == "userName"):
+            profile.userName = data 
+        elif (name == "preferredName"):
+            profile.preferredName = data
+        elif (name == "preferredPronouns"):
+            profile.preferredPronouns = data
+    
+        profile.save()
+        return JsonResponse({'message': f'{name} saved successfully'}, safe=False, status = 200)
```
JRSiwiecki, line 105: I think here, the `safe=False` parameter can be removed, as the `safe` parameter only needs to be set to false if we are sending objects that are not JSON serializable, but in this case, you are just returning a string in the message, so `safe = False` shouldn't be necessary (though I would still test it to see if I'm correct or not). 

Point being that I don't see why `safe = False` would be needed here but not on line 111, which is nearly identical to this one.

JRSiwiecki (changes requested):
Everything looks good for the most part, I just added some small suggestions. Re-ping me on discord when you get a chance to either make changes or provide feedback!

mesh/profiles/views.py
tdang2180: Yeah, you are right. I removed it 👍.

JRSiwiecki (approved):
Good work 👍🏼

Issue #118, opened 2023-04-30
F1.1 Theme Toggle Setting UX


Leaversa:
Story: As a night owl account owner mentor, mentee, or educator, I want to be able to switch to a dark theme version of the site so I can look at it better.

Details: Include a toggle button to enable/disable Dark or Light Theme in the User Settings page.

Technical Details:

- Front-end
- React
- HTML
- CSS
- UI
- UX

PR #292, opened 2023-08-22
Leaf/f1.1 theme toggle setting ux

Darren52BT:
Implements #118.

# What I've Done
I've created a switch button component that toggles the main theme between light and dark mode. The theme mode is also stored in local storage. If there is no theme mode in local storage, it will be initalized and stored as light mode. I have created a settings route to temporarily contain the toggle button, and I've added the corresponding link to the starting nav. 
I provided context for the theme, current theme mode, and the method used to toggle the theme from the top level, and that context is passed down from the ThemeContextProvider to its descendants, where it's used in the MainThemeProvider and the theme toggle component. 

# What You Should See
The added settings route
image
The Toggle Theme button at the Settings page
image
image
The theme mode being changed and stored in response to toggling
image
image


Leaversa (approved):
lgtm

PattyCakesxD (approved):
nice work

PR #293, opened 2023-08-23
Biography now saves to database.

Cynub:
Hello!

I've added a test case in "profiles_tests" that demonstrates the new functionality.

Whenever a user clicks "save" on their biography, the ProfileTextField's new prop, "handleSave" is called. This is passed down from the ProfileBiography component and is called on save.

I've also made it so that the ProfilePage component receives the 'accountID' prop, as it is necessary when making the post request to the server. As a result, the 'Profile' type has also been updated.

---------------------------------------------------------------------------------

While figuring out everything, I made some changes to the "accounts_tests" since there were errors showing. Eventually, I just stuck with "profiles_tests" but I left the changes here for review regardless.

This is my first pull request to a large project ever so please let me know if there's anything I should know!


mesh/profiles/views.py
```diff
@@ -26,20 +29,26 @@ def get(self, request, account_id, *args, **kwargs):
...
-        else:
-            return JsonResponse({'error': request.method + ' Method not allowed'}, status=405)
+        Updates the biography of the listed profile.
 
+        Returns a JSON response containing the saved biography of the specified profile through id.
+        """
+        data = json.loads(request.body)
+
+        if 'accountID' not in data:
+            return JsonResponse({'error': 'Invalid request. Missing accountID field.'}, status=401)
```
misslame, line 44: There is a folder in 'mesh' for the exceptions that we have. This doesn't look like this was your code but maybe we can open up a task for tech debt or you can take it on a part of this task to properly use the exceptions for invalid input.  I think I am leaning towards another tech debt task.

meshapp/src/profile/profile-page.tsx
```diff
@@ -295,6 +308,7 @@ const TestComponent = (props: any) => {
       placeholder={"Test"}
       text={"Test"}
       charLimit={200}
+      handleSave={() => {return}}
```
misslame, line 311: What is the purpose of this line? is this tied to testing? lol maybe I am dumb

meshapp/src/profile/profile-page.tsx
```diff
@@ -177,7 +181,15 @@ const ProfileOccupation = (props: {
  * @param props - Properties of the component
  * @param {string} props.biography - The initial text content of the bio
  */
-const ProfileBiography = (props: { biography: string }) => {
+const ProfileBiography = (props: { biography: string, accountID: number}) => {
+
+  function saveBiography(text: string) {
```
misslame, line 186: Good idea making a function.

misslame (approved):
Just Comments but overall self explanatory

meshapp/src/profile/profile-page.tsx
Cynub: I added a 'handleSave' function parameter to the parameters in 'profile-textfield.tsx'
Because of this, each <ProfileTextField> component is expected to be passed a 'handleSave' prop (since every editable field can be saved)

As such, the test component also needs one. The empty function is just there to satisfy the parameter requirement.

misslame: ahh makes sense. Interesting okay. Thanks for explaining to me.

tdang2180 (approved):
Everything seems fine, but I have a couple of suggestions. 
Great work! 🥇

mesh/profiles/views.py
```diff
@@ -26,20 +29,26 @@ def get(self, request, account_id, *args, **kwargs):
...
+        if 'biography' not in data:
+            return JsonResponse({'error': 'Invalid request. Missing biography field.'}, status=400)
+        else:
+            user = Profile.objects.get(accountID=data["accountID"])
```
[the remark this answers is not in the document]
Cynub: Will do!

mesh/tests/profiles_tests.py
```diff
@@ -101,4 +101,15 @@ def test_biography(self):
...
+    def test_biography_update(self):
+        """
+        Test Case for seeing if biography can be updated from specified account
+        
+        A POST request is sent to the '/profiles/biography' endpoint.
+        The test passes if the response status code is 200.
+        """
+        test_user = Account.objects.get(email="profilestest@gmail.com")
+        response = self.client.post(f"/profiles/biography",{'accountID': test_user.accountID, 'biography': "Testing..."}, content_type='application/json')
+        self.assertEquals(response.status_code, 200)
```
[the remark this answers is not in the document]
Cynub: Fixed!

mesh/profiles/views.py
Cynub: Yeah, that sounds like a tech-debt task 👍

PR #299, opened 2023-08-30
Hotfix: Profile Tests Missing accountID Attribute

Cynub:
Fix for an earlier PR/Merge where I updated the "profile" type to have an accountID attribute.

The example profiles used in the front-end compilation were not updated, and thus would cause compilation/build failures.

I've added the attribute to the tests as a fix while I figure out how/if I will refactor the code.

### DIFF
--- a/mesh/profiles/urls.py
+++ b/mesh/profiles/urls.py
@@ -19,8 +19,6 @@ urlpatterns = [
 
     # GET /profiles/user-name/:account_id
     # POST /profiles/user-name/:account_id
-    path("biography/<int:account_id>", profile_views.BiographyView.as_view()),
-    path("profile-picture/<int:account_id>", profile_views.ProfilePicturesView.as_view(), name="profile-picture"),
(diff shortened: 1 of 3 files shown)


misslame (approved):

mesh/profiles/urls.py
```diff
@@ -18,6 +19,8 @@
 
     # GET /profiles/user-name/:account_id
     # POST /profiles/user-name/:account_id
+    path("biography/<int:account_id>", profile_views.BiographyView.as_view()),
+    path("profile-picture/<int:account_id>", profile_views.ProfilePicturesView.as_view(), name="profile-picture"),
```
JRSiwiecki, line 23: What are these 2 lines doing here?

JRSiwiecki (commented):
Seems mostly good, just one small question.

mesh/profiles/urls.py
Cynub: Those are duplicate views. They don't need to be there. Thanks for catching that.

JRSiwiecki (approved):
Good fixes 👍🏼

Issue #32, opened 2023-03-30
A5.6 HTTP Post Call Handling


misslame:
As a User (Mentor, Mentee, or Educator), I want to be able to upload schools I attend(ed) and degrees or certificates I received to my profile

Details: Receive POST calls in order to store education changes to a user's profile. 

Technical Details:
- Back-end
- MySQL
- Django

PR #300, opened 2023-09-01
Leaf/a5.6 http post call handling

IanWearsHat:
[#32](https://github.com/LetsMesh/Site/issues/32) 

A POST request can be sent to /education/ endpoint.

Given a JSON body with accountID, degreeName, and collegeName, a new Education row will be created and connected with an EducationBridge row to the account specified by the accountID. An empty optionalDescription will be created if not specified in the request.

I created tests in education_tests.py for error handling and creating a successful POST request.

---

I don't remember in which issue I saw this comment, but it was suggested that the fields for EducationBridge be changed so that an account can have multiple educations. However, because this isn't the case for now, I did not create any checks for any duplicate education rows because trying to test for these raises an integrity error in the server.


mesh/urls.py
```diff
@@ -49,5 +50,6 @@
     path('settings/', include(accountsettings_urls)),
     path('confirmation/', include(confirmation_urls)),
     path('occupations/', include(occupations_urls)),
+    path('education/', include(education_urls)),
```
JRSiwiecki, line 53: Extremely small nitpick, but please makes this `educations/` for consistency.

mesh/education/views.py
```diff
@@ -1,3 +1,83 @@
...
+            return JsonResponse({'error': 'Invalid JSON format.'}, status=400)
+
+        except MissingRequiredFields:
+            return JsonResponse({'error': 'Missing required JSON fields.'}, status=400)
+
+        except ValueError:
+            return JsonResponse({'error': 'A field has invalid type.'}, status=400)
+
+        except (Account.DoesNotExist, Profile.DoesNotExist):
+            return JsonResponse({'error': 'Account or Profile not found.'}, status=400)
```
JRSiwiecki, line 70: This status code should likely be `404 Not Found` rather than `400 Bad Request`.

JRSiwiecki (changes requested):
There are some small things that should be changed, so re-ping me on discord when you get a chance to review my feedback.

Also, since I believe the Education/EducationBridge model updates were finished while you were working on this, I would like to let this be merged, HOWEVER, if possible I think you should try to pull those changes into your branch and update the POST handling to have the intended functionality of no duplicate Educations, and just create new EducationBridges when needed (or rather, similar to how Occupation and OccupationBridge are implemented). Currently, if these changes are merged, I'm pretty certain they will not work since they will be accessing parts of Education and EducationBridge that have been switched around.

GuyWhoCode (approved):
LGTM

JRSiwiecki (approved):
👍🏼 good job! Also in the future, add a keyword like "closes" or "resolves" in front of the issue number in the description (though this might not matter because supposedly this only works when merging into the "default" branch) so that Github can automatically link the issue.

You can also try linking the issue in the sidebar to the right on the conversation page.

PR #302, opened 2023-09-02
Re-name settings url to account-settings to match folder name

### LINKED ISSUE #301: BUG: Re-name settings url to account-settings
In the mesh folder, the urls.py contains all the base url paths for the website.

The settings URL path could be renamed to `account-settings` to be more clear, as well as to match up with the folder name for its corresponding model, `accountSettings`.

Note: To set/follow standards, I would recommend the url path be named `/account-settings/` rather than `/accountSettings/`

JRSiwiecki:
Yup that's it

Resolves #301


chrispybokchoy (approved):
looks good

BradJKim (approved):

PR #306, opened 2023-09-10
A3.1 name nickname and pronouns

hcnguyen1:
Users are able to change their name and display their nicknames and pronouns

### DIFF
--- a/meshapp/src/profile/profile-name.tsx
+++ b/meshapp/src/profile/profile-name.tsx
@@ -1,0 +1,138 @@
+import React, { useState } from "react";
+import { TextField, Box, ThemeProvider, Theme, Typography, Grid, createTheme } from "@mui/material";
+
+import EditIcon from "@mui/icons-material/Edit";
+import SaveIcon from "@mui/icons-material/Save";
+
+
+/**
+ * A React component that renders a text field with editing capabilities.
+ * Includes a character limit and an edit/save mode toggle.
+ *
+ * Used in the Profile page (src/profile/profile-page.tsx).
+ *
+ * @param props - Properties of the component
+ * @param {string} props.label - The label
+ * @param {string} props.placeholder - The placeholder text
+ * @param {string} props.text - The initial text content
+ * @param {number} props.charLimit - The max number of characters allowed
+ */
+const theme = createTheme({
+    palette: {
+      mode: "light",
+      primary: {
+        main: "#0b7d66",
+      },
+    },
+    typography: {
+      h1: {
+        fontFamily: "cocogoose",
+        fontWeight: "bold",
+        color: "#26383A",
+      },
+    },
+});
+
+const ProfileName = (props: {
+  label: string;
+  placeholder: string;
+  text: string;
+  charLimit: number;
+}) => {
+  const [text, setText] = useState(props.text);
+  const [editMode, setEditMode] = useState(false);
+
+  // Enforce developer-defined character limit
+  const handleTextChange = (event: any) => {
+    if (event.target.value.length > props.charLimit) return;
+    setText(event.target.value);
+  };
+
+  const handleEditClick = () => {
+    setEditMode(!editMode);
+  };
+
+  const handleSaveClick = () => {
+    setEditMode(false);
+  };
+
+  return (
+
+    // Bug: Recent user changes does not change a user's username
+    
+    editMode ? (
+        <ThemeProvider theme={theme}>
+            <Box sx={{ pl: "40px", display: "flex", alignItems: "flex-end" }}>
+                <Typography
+                    variant="h1"
+                    sx={{
+                        lineHeight: 1,
+                        display: "inline",
+                        fontSize: "60px",
+                    }}
+                >
+                    <TextField
+                        required
+                        id="standard-required"
+                        label="Username"
+                        defaultValue={props.text}
+                        variant="standard"
+                        onChange={handleTextChange}
+                    />
+                    <SaveIcon
+                        color="primary"
+                        onClick={handleSaveClick}
+                        sx={{
+                            "&:hover": {
+                            color: "#0A6B57",
+                            },
+                            cursor: "pointer",
+                            transition: "color 0.15s ease-in-out",
+                        }}
+              />
+                </Typography>
+            </Box>
+        </ThemeProvider>
+        
+    
+    ) : (
+        <ThemeProvider theme={theme}>
+        <Box sx={{ pl: "40px", display: "flex", alignItems: "flex-end" }}>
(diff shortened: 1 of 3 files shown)


PattyCakesxD (changes requested):
When trying to update the nickname, the name auto-truncates to the first word it gets if there's spaces included. Like, if I try to put in "**Joe Biden**", it automatically truncates it to just "**Joe**".  Something similar, if you try to save an empty nickname, it auto-reverts to whatever the previous name was (if this is intended, nice - just an observation)

Also install the Prettier extension on vscode so the code looks cleaner

meshapp/src/profile/profile-name.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const [text, setText] = useState(props.text);
+  const [editMode, setEditMode] = useState(false);
+  
+
+  // Enforce developer-defined character limit and empty string
+  const handleTextChange = (event: any) => {
+    if (event.target.value.length > props.charLimit) return;
+    if (event.target.value.trim() === "") return;
+    if (!/^[a-zA-Z\s]+$/.test(event.target.value)) return;
+    setText(event.target.value);
```
misslame, line 53: For all three of these, there should be a little more than just a return. Ideally, you would tell the user what was wrong.

meshapp/src/profile/profile-name.tsx
```diff
@@ -0,0 +1,157 @@
...
+  let nicknameOrPronouns = "Nickname";
+
+  const nameOrPronounFont = (charLimit: number) => {
+    if(props.charLimit == 15) {
+      return name;
+    } else {
+      nicknameOrPronouns = "Pronouns"
+      return pronoun;
+    }
+  }
```
misslame, line 69: I am not sure I follow the logic here despite the comment. Can I ask what this is supposed to do?  What is "nameOrPronounFont"

misslame (changes requested):
Some small comments. Overall, looks good, just needs improvements on error handling

meshapp/src/profile/profile-name.tsx
```diff
@@ -0,0 +1,157 @@
...
+}) => {
+  const [text, setText] = useState(props.text);
+  const [editMode, setEditMode] = useState(false);
+  
+
+  // Enforce developer-defined character limit and empty string
+  const handleTextChange = (event: any) => {
+    if (event.target.value.length > props.charLimit) return;
+    if (event.target.value.trim() === "") return;
+    if (!/^[a-zA-Z\s]+$/.test(event.target.value)) return;
```
rezatnoMsirhC, line 52: 1. I think you should handle all of the validations with a single regex:
   1. Character-limit can be imposed via regex
      1.  For example, `\w{1,32}` matches any string of length 1 to 32 of alphanumeric and underscore characters, like `ABCD_EFG`
   1. Checking if the string is just whitespace can be accomplished with regex
      1. For example, `\w{1,16} \w{1,16}` would not match a string of just whitespace, but would allow something like `A B`
1. As-is, this function would allow something like `H E L L O       W O R L D` to pass. Is that something we want? Also do we not want to allow numbers in this text?


*btw I recommend using [RegExr](https://regexr.com) for messing around with regex*

meshapp/src/profile/profile-name.tsx
```diff
@@ -0,0 +1,157 @@
...
+  };
+
+  // Checks if it is a name or pronoun
+  
+  const name = "60px";
+  const pronoun = "30px";
+  let nicknameOrPronouns = "Nickname";
+
+  const nameOrPronounFont = (charLimit: number) => {
+    if(props.charLimit == 15) {
```
rezatnoMsirhC, line 63: Avoid magic numbers:
1. What does `15` mean in this context? Define a constant with a clear meaning.
2. This is especially dangerous/error-prone if you decide to change wherever else you use 15, and forget to change it here too. Use the constant in all applicable places.

meshapp/src/profile/profile-name.tsx
```diff
@@ -0,0 +1,157 @@
...
+    setText(event.target.value);
+  };
+
+  // Checks if it is a name or pronoun
+  
+  const name = "60px";
+  const pronoun = "30px";
+  let nicknameOrPronouns = "Nickname";
+
+  const nameOrPronounFont = (charLimit: number) => {
```
rezatnoMsirhC, line 62: imo this is a weird way to accomplish polymorphism. I think the intent is to have something like `NameTextField` and `PronounTextField` with their own specific font-size. I suggest defining a polymorphic class hierarchy so that `NameTextField` and `PronounTextField` extend/are `TextField`, and they can do/be all the things `TextField` can do, but possibly in their own special ways. Maybe right now this is just for font-size, but maybe later we do stuff like have each `TextField` implement its own validation scheme.

meshapp/src/profile/profile-name.tsx
hcnguyen1: Used Snackbar to tell user what's wrong.

testing comment

meshapp/src/profile/profile-name.tsx
hcnguyen1: Intended to choose between "Nickname" or "Pronoun" depending on font size.
Hard to read so I adjust the code to using parameters instead.

meshapp/src/profile/profile-name.tsx
hcnguyen1: Code Removed

meshapp/src/profile/profile-name.tsx
hcnguyen1: Removed code.
Created a new way such as making the parameter of the class for font size.

meshapp/src/profile/profile-name.tsx
hcnguyen1: Handled with single regex.

meshapp/src/profile/profile-page.tsx
```diff
@@ -51,7 +56,10 @@ const ProfilePage = (props: Profile) => {
   return (
     <Box className="profile-page-container">
       <Box className="profile-page-header">
-        <ProfileHeader name={props.name} pronouns={props.pronouns} />
+        
+        {/* <ProfileHeader name={props.name} pronouns={props.pronouns} /> Temporary Comment*/}
```
rezatnoMsirhC, line 60: What's this for?

hcnguyen1: Old ProfileHeader Deleted. 
Renamed the class from ProfileName to ProfileHeader

rezatnoMsirhC (approved):

misslame (approved):

PattyCakesxD (approved):

PR #308, opened 2023-09-12
A5.5 http get call handling Education

### LINKED ISSUE #31: A5.5 HTTP Get Call Handling
As a User (Mentor, Mentee, or Educator), I want to be able to upload schools I attend(ed) and degrees or certificates I received to my profile

Details: Allow for the fetching of the User's education through GET API Call in order to display the name 

Technical Details: 
- Back-end
- MySQL
- Django

IanWearsHat:
Resolves #31 

Made GET call for getting all education rows in the entire table.
Made GET call for getting all education rows for a specific account.

### DIFF
--- a/mesh/education/urls.py
+++ b/mesh/education/urls.py
@@ -3,6 +3,7 @@ from django.urls import path
 from . import views as education_views
 
 urlpatterns = [
-    # POST  /education/
+    # POST  /educations/
+    # GET   /educations/
     path('', education_views.EducationView.as_view(), name="education")
 ]
--- a/mesh/education/urls.py
+++ b/mesh/education/urls.py
@@ -5,5 +5,9 @@ from . import views as education_views
 urlpatterns = [
     # POST  /educations/
     # GET   /educations/
-    path('', education_views.EducationView.as_view(), name="education")
+    path('', education_views.EducationView.as_view(), name="education"),
+
+    # GET   /educations/:account_id
+    path("<int:account_id>/", education_views.EducationsDetailView.as_view(),
+    name="user_educations")
 ]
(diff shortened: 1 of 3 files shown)


Cynub (approved):
Looks good to me (no formatting issues, tests pass) 👍

mesh/education/urls.py
```diff
@@ -3,6 +3,11 @@
 from . import views as education_views
 
 urlpatterns = [
-    # POST  /education/
-    path('', education_views.EducationView.as_view(), name="education")
+    # POST  /educations/
+    # GET   /educations/
+    path('', education_views.EducationView.as_view(), name="education"),
+
+    # GET   /educations/:account_id
```
JRSiwiecki, line 10: Extremely small nitpick, but I believe I was just helping Brad with this issue, but clarify in this comment that the endpoint is /educations/:account_id/ with a trailing slash, as we were having problems where making requests to an endpoint that required a slash WITHOUT a slash, would not work correctly. So I would keep the trailing slash, just update the comment to match. Possibly a similar situation with the other two endpoints at /educations/

mesh/education/views.py
```diff
@@ -14,17 +15,32 @@
...
+        Handle GET requests to /educations/ endpoint.
+
+        Returns a JSON response containing all Educations.
+        """
+        educations = Education.objects.all()
+
+        # Create a list of dictionaries containing education details
+        educations_list = serializers.serialize("json", educations)
+
+        return HttpResponse(educations_list, status=200)
```
JRSiwiecki, line 29: Good way to deal with the issue you were mentioning. 👍🏼

Cynub: Right? An HttpResponse instead of a JSON Response. +100000 points.

JRSiwiecki (approved):
Good work!

mesh/education/urls.py
IanWearsHat: I literally had this problem the first time I tried using Postman for this project. I will fix.

Issue #310, opened 2023-09-19
BUG: API accounts is not performing as expected


bluesimp1102:
The request to accounts API paths is not performing correctly after the changes in the Account model (encrypting password added)

Relate to: [#67 B1.2 HTTP Post Call Handling](https://github.com/orgs/LetsMesh/projects/2?pane=issue&itemId=28684906)

bluesimp1102:
PR: #311[https://github.com/LetsMesh/Site/pull/311]

PR #311, opened 2023-09-19
bugfix #310

bluesimp1102:
fix the request handlings for create, update account, get one or all the accounts, and the accounts test

Issues: 
- Bug issue: [Site #310](https://github.com/orgs/LetsMesh/projects/2/views/1?pane=issue&itemId=39055351)


mesh/accounts/views.py
```diff
@@ -40,38 +42,71 @@ def post(self, request, *args, **kwargs):
...
+        serialized_data = serialize('json', [account])
+        parsed_data = json.loads(serialized_data)
+
+        # Extract the first element from the parsed list
+        single_account_data = parsed_data[0] if parsed_data else {}
+
+        return HttpResponse(
+            json.dumps(single_account_data), 
+            status=200,
+            content_type='application/json'
```
JRSiwiecki, line 107: Why does this have to be serialized, loaded from json, and then converted back to json?

mesh/accounts/models.py
```diff
@@ -7,15 +7,8 @@ class Account(models.Model):
...
-    salt = models.BinaryField(max_length = 36,editable = True) 
-    ###
+    salt = models.BinaryField(max_length = 36,editable = True)
     phoneNum = models.CharField(max_length=15)
-    DISPLAY_THEMES = [
-        ('L', 'Light'),
-        ('D', 'Dark'),
-    ]
-    displayTheme = models.CharField(max_length=1, choices=DISPLAY_THEMES, default='L')
-    enabled2Factor = models.BooleanField(default=False)
```
JRSiwiecki, line 18: Just to confirm, were these properties moved to some other table?

JRSiwiecki (commented):
Everything looks good for the most part, just some small comments. My apologies for taking so long to review, it slipped my mind for the most part.

Re-ping me on discord after you get a chance to check out my comments.

mesh/accounts/models.py
bluesimp1102: yes it has moved to the accountSettings model

mesh/accounts/views.py
bluesimp1102: as far as I know, Django's ORM returns you a model instances, which are not directly serializable into JSON format, so I serialized it to a json string, then load it to a json object then dumps it back to reconvert the Python dictionary to a JSON string, which suitable for an HTTP response. I tried to do it without the complication you mentioned but the response either gets messy or wouldn't return expected result

misslame (approved):

GuyWhoCode (approved):
LGTM

bluesimp1102:
- modified tests relating to the changes in Accounts model
- added API paths for AccountSettings to create and update account settings, fixed path to get displayTheme (from `/displayTheme` to `/displayTheme/<int:account_id>`
- added tests to create/update account settings

PR #312, opened 2023-09-19
Leaf/b3.1 2 factor account setting

### LINKED ISSUE #75: B3.1 2Factor Account Setting
Story: **As a User (Mentor, Mentee, or Educator), I want to be able to have a more secure login where it requires another device of mine in order to login**

Details: Allow for the user to opt-in through their account settings for 2 factor authentication. Ensure that this is a persistent account setting. Do not use client-based storage. 

Technical Details: 
- React
- Fullstack
- SQL
- Databases
- HTTP Rest Calls (GET/POST)
- Django

BradJKim:
Settings page with setting options for the user account.

- Toggle settings for user account (currently only 2FactAuth)
- Saves settings options and is displayed when settings page is built 

Todo:

- Need to confirm 2FactAuth Settings with email
- Styling is basic, may want to create better visual aesthetics with color and spacing
- New Toggle components needed for other settings with non-boolean values

### DIFF
--- a/mesh/accountSettings/urls.py
+++ b/mesh/accountSettings/urls.py
@@ -3,5 +3,8 @@ from . import views as accountsettings_views
 
 urlpatterns = [
     path("displayTheme", accountsettings_views.display_theme, name="display_theme"),
-    path("settings", accountsettings_views.showSettings.as_view(), name="settings")
+
+    # GET    /api/settings/:account_id
+    # PATCH  /api/settings/:account_id
+    path("settings/<int:account_id>", accountsettings_views.showSettings.as_view(), name="settings")
 ]
--- a/mesh/accountSettings/urls.py
+++ b/mesh/accountSettings/urls.py
@@ -2,5 +2,6 @@ from django.urls import path
 from . import views as accountsettings_views
 
 urlpatterns = [
-    path("displayTheme", accountsettings_views.display_theme, name="display_theme")
+    path("displayTheme", accountsettings_views.display_theme, name="display_theme"),
+    path("twoFactAuth", accountsettings_views.two_fact_auth, name="two_fact_auth")
 ]
--- a/mesh/accountSettings/urls.py
+++ b/mesh/accountSettings/urls.py
@@ -3,5 +3,5 @@ from . import views as accountsettings_views
 
 urlpatterns = [
     path("displayTheme", accountsettings_views.display_theme, name="display_theme"),
-    path("twoFactAuth", accountsettings_views.two_fact_auth, name="two_fact_auth")
+    path("twoFactAuth", accountsettings_views.TwoFactAuthView.two_fact_auth, name="two_fact_auth")
 ]
--- a/mesh/accountSettings/urls.py
+++ b/mesh/accountSettings/urls.py
@@ -3,5 +3,5 @@ from . import views as accountsettings_views
 
 urlpatterns = [
     path("displayTheme", accountsettings_views.display_theme, name="display_theme"),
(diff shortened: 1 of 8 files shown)


mesh/accountSettings/urls.py
```diff
@@ -2,5 +2,9 @@
 from . import views as accountsettings_views
 
 urlpatterns = [
-    path("displayTheme", accountsettings_views.display_theme, name="display_theme")
+    path("displayTheme", accountsettings_views.display_theme, name="display_theme"),
+
+    # GET    /api/settings/:account_id
+    # PATCH  /api/settings/:account_id
```
JRSiwiecki, line 8: Might want to update these comments, as I don't believe this is the correct path begins with api/... for these requests.

mesh/accountSettings/urls.py
```diff
@@ -2,5 +2,9 @@
...
+    path("settings/<int:account_id>", accountsettings_views.showSettings.as_view(), name="settings")
```
JRSiwiecki, line 9: Just checking, was this the path it had when we were working through those weird trailing slash issues? If the endpoint works fine then disregard this, I'm just curious.

mesh/accountSettings/views.py
```diff
@@ -29,3 +33,52 @@ def display_theme(request):
...
+        """ 
+        Handles GET request
+
+        Return JSON with Two Factor Authentication Settings for given account ID. 
+        Returns a 404 error with error message if account does not exist.
+        """
+        try:
+            settings = Settings.objects.get(accountID=account_id)
+            settings_detail = serializers.serialize('json', [settings])
+            return JsonResponse(settings_detail, safe=False)
```
JRSiwiecki, line 52: I think here, instead of using JsonResponse, use `HttpResponse(settings_detail, content_type="application/json")` so that we don't double serialize our response data, as someone brought this up a week or two ago and it makes sense to just return it as HttpResponse instead.

meshapp/src/Settings/settings-page.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  /** 
+   *  NOTE: Second useEffect for sending patch request to update account settings each time settings is modified.
+   *        The patch request is made every time the setting switch is changed.
+   *        This may want to be changed in the future to ensure that only one 
+   *        patch request is made for multiple settings.
+   */
+  useEffect(() => {
+    // Make a PATCH request to the backend API to update account settings 
+    axiosInstance.patch("settings/settings/" + props.accountID + "/", {...settings}) // NOTE: Use accountSettings api name when merging
```
JRSiwiecki, line 97: Same comment as above.

meshapp/src/Settings/settings-page.tsx
```diff
@@ -0,0 +1,129 @@
...
+        setLoading(false);
+      })
+      .catch((error) => {
+        console.error('Error patching account settings:', error);
+        setLoading(true);
+      });
+  }, [settings])
+
+  const handleToggleChange = (settingName: any) => (event: any) => {
+    setSettings({ ...settings, [settingName]: event.target.checked }); // TODO: Implement Confirm Authentication for 2FactAuth
```
JRSiwiecki, line 108: For TODOs, I would suggest to write up an issue explaining what needs to be done, and mention this PR in that issue.

JRSiwiecki (changes requested):
I left some comments and some questions, re-ping me on discord when you get a chance to look them over!

P.S. I would suggest getting someone from the front-end or full-stack team to review your code as well, so that that they can review the React code, as they would probably do better than I would! Also make sure that when you are pinging people on discord for reviews, you also add them as a reviewer on Github.

Issue #77, opened 2023-04-19
B3.3 Confirm 2Factor Authentication


misslame:
Story: **As a User (Mentor, Mentee, or Educator), I want to be able to have a more secure login where it requires another device of mine in order to login**

Details: Ensure that the user can login through 2factor authentication. When a user logs in, generate a random code, email the code, prompt for the code and then authenticate said code after input.

Technical Details: 
- Full-stack
- React
- Django
- Generation
- Basic Authentication.

PR #316, opened 2023-10-17
Leaf/b3.3 confirm 2 factor authentication

gregoriusavip:
Solving issue: #77 
# Features
- Allow user to log in as long as the account is on the database
- Upon logging in, the user will now be asked to confirm their 2FA if they have it enabled

# Implementation
- Added an extra field called otp_base32. This will hold the seed of the user's randomly generated base 32 for their One Time Password (OTP)
- Added a services.py file to handle function services related to the account view
- For 2FA generation/verification, the package **pyotp** is used. The package also allows this implementation to be further extended for MFA implementation.
- Added 3 Views: Set2FAView, Verify2FAView, and LoginView
- Created a basic web page to input the 6-digit (OTP) at \meshapp\src\home\otp.tsx. The input field relies on a package: **react-otp-input**
- To transfer some needed data from page to the other, the package: **react-cookie** is used, which implements the cookie functionality

# How It Should Work

1. The user first will make a login request from the front end, filling in the required field of both email and password
2. Upon clicking the login button, a post request to the login view from the backend is made
3. Backend will verify if the request is valid, verifying if the user's data and password. If successful, backend will give both the accountID that ties to this user and the information if the user has 2FA enabled
4. If 2FA is enabled, the page will be navigated to /otp where the user will be asked to input a 6 digit pin code, otherwise they will be directed to the logged_in_home page
5. Simultaneously, a cookie holding the user's accountID will be created with a time limit. It will also call another post request to the backend, calling Set2FAView.
6. Set2FAView will check if the user has 2FA enabled or not. If it doesn't then backend will generate a new seed and store it on the otp_base32 field for that user. Finally, backend will send an email to the corresponding user's email which holds the otp. **NOTE** Set2FAView might need a separate new view or function to handle the user's first-time setup, especially if MFA uses an authenticator to be implemented.
7. Once the user input the 6 digit code and click the submit button, a post request is called for Verify2FAView, which take the cookie holding the user's accountID to find the correct OTP seed, and then verify if the 6 digit code is correct.
8. Upon verification, the user will be redirected to the logged_in_home page.

# Issues/Bugs/ToDos

- [x] ~~BUG: upon first login request and the cookie was generated, there is a synchronization error between the cookie data and the request. This causes the user to be verified and their user ID is still saved as a cookie, but upon requesting OTP verification, the user's ID is somehow still undefined. This does not occur on a subsequent request, and the possible explanation is that the previous cookie from a previous attempt is still saved.~~ SOLVED: 3d188d8
- [x] ~~There needs to be a timeout when the user is on the otp page, since the otp that is sent to the user will expire on a given period of time. On this point, it might also be needed for the otp that was sent to the user to be semi-constantly checked until it no longer valid on the backend, and then synchronized with the timeout on the front end, guaranteeing a possible better consistency for the otp period.~~ Somewhat solved by having session interval: 82aa946
- [ ] Email from backend to the user has not been setup and therefore currently the otp is printed out to the console
- [ ] While the code for sending the email is there, the body message for the email itself need to be further extended
- [ ] In order to organize the files, most of the new implemented backend functionality is stored at services.py. But to verify user's login request, the decrypt function needs to be copied to this file. To avoid copying of the same code, it is possible to move some functions from views.py to services.py such as both encrypt and decrypt.

# Improvement
- [ ] Permission/authentication needs to be added for all of the newly added views. For example, requesting an OTP verification does not make sense if the user is not verified to be logged in.
- [ ] Similarly, the /otp web page should not be accessible if the user is not properly verified
- [ ] Some extra security need to be implemented. Some detail for the basic security requirement: https://pypi.org/project/pyotp/. For example, login requests should be throttled in order to avoid brute-force attack
- [ ] The web page implementation is pretty barebones. Some more styling can be added such as handling error responses with better UI instead of just displaying an error alert

# Note
While packages added to the project frontend are saved on package.json, I am not sure if that is also the case for the backend. Therefore to test the changes these packages needed to be installed:

- pyotp
- django-cors-headers

django-cors-headers was used for testing calls made from front end to the back end. To do this, I run the front end using **npm start** and the backend using **manage.py runserver**. At the same time on django settings, both field, CORS_ORIGIN_WHITELIST and CORS_ALLOWED_ORIGINS are set as my localhost for the front end, which is "http://localhost:3000"


mesh/settings.py
```diff
@@ -62,6 +64,7 @@
 ]
 
 CORS_ORIGIN_WHITELIST = [
+    #allow request from localhost port 3000 for testing
```
misslame, line 67: What are these comments for?

meshapp/src/components/login-form.tsx
```diff
@@ -22,6 +25,52 @@ const LoginScreen = (props: ComponentProps) => {
...
+        }
+        else{
+          navigate('/logged_in_home')
+        }
+      })
+      .catch((error) => {
+        if(error.response){
+          // The request was made and the server responded with a status code
+          // that falls out of the range of 2xx
+          alert(error.response.data.message); //TODO: Improve the error response
```
misslame, line 60: Can you make an issue for this TODO

meshapp/src/components/login-form.tsx
```diff
@@ -22,6 +25,52 @@ const LoginScreen = (props: ComponentProps) => {
...
+          .catch((error) => {
+            alert(error.response.data.message);
+          });
+        }
+        else{
+          navigate('/logged_in_home')
+        }
+      })
+      .catch((error) => {
+        if(error.response){
```
misslame, line 57: For error handling, it might be worth to use more descriptive erroring potentially included with the default error responses.

meshapp/src/home/otp.tsx
```diff
@@ -0,0 +1,143 @@
...
+          height: "4.5rem",
+          fontSize: "1.5rem",
+          border: '2px solid #ccc',
+        },
+      },
+    },
+  },
+});
+
+const Otp = () => {
```
misslame, line 33: maybe this is a dumb question what does OTP mean

misslame (approved):

misslame (changes requested):
Oops meant to do request changes

mesh/settings.py
gregoriusavip: If I understand correctly when I launched both the backend and frontend at the same time locally, the frontend would have a different domain. Adding the CORS_ORIGIN_WHITELIST allowed the frontend to request to the backend. 

That comment was just a reminder that the backend currently allows requests from that port and I'm not too sure if this needs to be removed eventually.

This is based on a Django extension from [here](https://github.com/adamchainz/django-cors-headers) although the configuration setting is now called CORS_ALLOWED_ORIGINS

meshapp/src/home/otp.tsx
gregoriusavip: One-Time Password, the pin that the user inputs for 2FA/MFA. Is that what you are asking?

meshapp/src/components/login-form.tsx
gregoriusavip: I extended the error handler. But I wasn't sure if the error description was for the user or the console, so for now I only extended it for the console. [fc1474c](https://github.com/LetsMesh/Site/pull/316/commits/fc1474c655e43ab86f792d0326b90d3133981852)

meshapp/src/components/login-form.tsx
gregoriusavip: I was going to make an issue for this but I thought since it wasn't that complicated I went ahead and added sweetalert so now it looks like this for errors:
image
And this is just a quick pop-up when the user is verified via OTP
image

[5a261af](https://github.com/LetsMesh/Site/pull/316/commits/5a261af9968c62e4f505966aa3493a49409b2916)

mesh/settings.py
misslame: Ah yea it just seemed redundant to me since just below it it says "localhost:3000"

meshapp/src/components/login-form.tsx
misslame: Can you remove the comment for the TODO if you're satisfied with how you solved it otherwise yea make an issue for later addressing (or closing) don't want to have todo statements in code without issues attached

meshapp/src/components/login-form.tsx
misslame: Yea it is for the console but currently it spits out the default exception. I think proper console logging / error logging would include a more "english" error statement before outputting the exception. This is something nice to have not necessarily required

meshapp/src/home/otp.tsx
misslame: Ahh can we have a comment where "otp" is first introduced.

meshapp/src/components/login-form.tsx
gregoriusavip: yeah I believe it should be removed on the latest commit

meshapp/src/components/login-form.tsx
gregoriusavip: This is how it currently looks like for console error. https://github.com/LetsMesh/Site/blob/82aa9466cabd90755fe24d0b4a3da7e758675390/meshapp/src/config/errorHandlerModule.ts#L10-L33
Would that be fine? Also, I'd assume that the comments would be redundant here

meshapp/src/home/otp.tsx
gregoriusavip: Would this be fine? bf483af

mesh/settings.py
gregoriusavip: its removed. 3e2025d

mesh/accounts/services.py
```diff
@@ -0,0 +1,80 @@
...
+from django.http import JsonResponse
+from django.core.mail import send_mail
+import pyotp, bcrypt, json, os
+
+def decrypt(password, salt):
+    pepper = os.getenv("PEPPER")
+    password = f"{password}{pepper}".encode('utf-8')
+    return bcrypt.hashpw(password,salt)
+
+def getUserServices(request):
```
JRSiwiecki, line 11: This is somewhat nitpicky, but python functions should be named with snake_case. This would also be good so that this code can be consistent with the rest of the codebase. Another small thing would be to use a formatter like autopep8 or black, though since most of the codebase doesn't have some formatter set in place that isn't as much of an issue for this PR specifically (though it would be nice to try and set that standard now!)

mesh/accounts/services.py
```diff
@@ -0,0 +1,80 @@
...
+    totp = pyotp.TOTP(otp_base32, interval=60) #for debug, remove after sending email works
+    #sending email need authentication
+    # send_mail(     
+    #     "User OTP",
+    #     "the otp: {}".format(pyotp.TOTP(otp_base32, interval=90).now),
+    #     os.environ.get("EMAIL_NAME"),
+    #     [user.email],
+    #     fail_silently=False,
+    # )
+    print(totp.now())   #for debug. remove after sending email works
```
JRSiwiecki, line 52: For parts like these, if it is not being handled in this PR then an issue should be created. I saw that you mentioned it in the PR description which is good but it would be more clear if there were issues, and I would write in those issues what should be removed just in case people don't stumble across these comments.

meshapp/src/components/login-form.tsx
```diff
@@ -22,6 +26,35 @@ const LoginScreen = (props: ComponentProps) => {
     console.log(formData);
   };
 
+  const navigate = useNavigate();
+  const [cookies, setCookie] = useCookies(['user_id']);
+
+  useEffect(() =>{},[cookies]);
```
JRSiwiecki, line 32: I think this line could use a comment explaining why the body for useEffect is empty, assuming it is for a good reason.

mesh/accounts/services.py
gregoriusavip: Should the issue be made before or after the PR is merged? I was assuming that since the changes haven't been approved yet I shouldn't make any issue if it's related to in-progress changes.

meshapp/src/config/errorHandlerModule.ts
JRSiwiecki, line 1: I like the errorHandler, very nice to give visual cues to the user when a problem occurs.

meshapp/src/home/otp.tsx
```diff
@@ -0,0 +1,212 @@
...
+  const [otp, setOtp] = useState(''); // inital OTP as empty
+  const navigate = useNavigate();
+  const [cookies, , removeCookie] = useCookies(['user_id']);
+
+  useEffect(() =>{},[cookies]);
```
JRSiwiecki, line 40: Similar to before, it would be nice if this line would have a comment.

JRSiwiecki (changes requested):
Other than style changes and some parts that could use some clarification, I don't have too many issues, after you've gotten a chance to respond to all my comments, re-ping me on discord and I'll check it out again. 

Good work 👍🏼 and very nice PR description as well!

mesh/accounts/services.py
JRSiwiecki: You're correct, it probably should be made after your work is merged. Link this PR to the issues you may create so that people can read the description to know what they should be looking to achieve.

meshapp/src/components/login-form.tsx
```diff
@@ -22,6 +26,35 @@ const LoginScreen = (props: ComponentProps) => {
...
+            navigate('/otp');
+          })
+          .catch(errorHandler);  // Somehow, user's account ID could not be found
+        }
+        else{
+          navigate('/logged_in_home')
+        }
+      })
+      .catch(errorHandler);  // User attempted to login and an error occured
+  }
```
JRSiwiecki, line 56: Looking back on these lines of code, I wonder if using try/catch along with await in more places instead of .then, .catch would be a bit more readable and doesn't require directly responding to promises.

meshapp/src/home/otp.tsx
```diff
@@ -0,0 +1,212 @@
...
+          icon: "success",
+          title: "Verified!",
+          showConfirmButton: false,
+          timer: 1500
+        });
+        removeCookie("user_id");
+        navigate('/logged_in_home');
+      })
+      .catch(errorHandler); // User's OTP could not be verified
+  }
```
JRSiwiecki, line 137: I would figure that something similar as to what I mentioned earlier about try/catch blocks instead of callbacks with .then and .catch could be used here.

meshapp/src/components/login-form.tsx
JRSiwiecki: Good job!

meshapp/src/home/otp.tsx
JRSiwiecki: 👍🏼

meshapp/src/home/otp.tsx
gregoriusavip: should be handled here 68b162a

meshapp/src/components/login-form.tsx
gregoriusavip: 68b162a

meshapp/src/components/login-form.tsx
misslame: I think this is fine but I was more leaning towards specific error outputting. Like imagine putting in a password and instead of saying "incorrect password" the error message is just "oops something went wrong" or an exception stack strace. 

Specific error handling will tell you what you did wrong, even in console log so that if someone is debugging an exception they have clues. 

I like to add a specific english explanation for the specific error and then include stack trace afterwards.

meshapp/src/home/otp.tsx
misslame: I was saying to define what the acronym "otp" is

mesh/accounts/services.py
gregoriusavip: I reformatted using black and also renamed the functions to snake_case c6522ce

mesh/accounts/models.py
```diff
@@ -18,6 +18,7 @@ class Account(models.Model):
     enabled2Factor = models.BooleanField(default=False)
     isMentor = models.BooleanField(default=False)
     isMentee = models.BooleanField(default=False)
+    otp_base32 =  models.CharField(max_length = 200, null = True)
```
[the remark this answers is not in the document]
gregoriusavip: I tried to add just enough for each class/function. Let me know if it's currently too much/too little. d63f172

meshapp/src/home/otp.tsx
gregoriusavip: Oh sorry. I added https://github.com/LetsMesh/Site/pull/316#discussion_r1525776503

mesh/accounts/services.py
JRSiwiecki: 👍🏼

mesh/accounts/models.py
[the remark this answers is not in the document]
JRSiwiecki: I think it's enough, good work!

meshapp/src/components/login-form.tsx
gregoriusavip: I made some changes. Let me know if anything is wrong. Also, I accidentally formatted login-form.tsx with prettier so it might've looked like many lines changed. c9c2d03

misslame (approved):

misslame:
Good work!!

Issue #80, opened 2023-04-25
B5.2 HTTP Post Call Handling


misslame:
Story: **As an account owner, I want to be able to edit account settings and information such as email, password, enable matching, etc.**

Details: When a save button is pressed on the client for shallow (small, toggle based, basic string, etc.) settings, the POST call should be handled with the settings changed. THIS DOES NOT INCLUDE EMAIL, 3rd PARTY ACCOUNT, or PASSWORD CHANGES. 

Technical Details: 
- Back-end
- Django
- MySQL
- Python
- Databases
- HTTP REST Post

PR #318, opened 2023-12-21
Leaf/B5.2 HTTP Post Call Handling #80

bluesimp1102:
Issue: #80 

- moved password changing operation to route **PATCH** `/accounts/change-password/`
- rename **POST** `/accounts/checkAccount` to **POST** `/accounts/check-password/`
- route **PATCH** `/accounts/:account_id/` now only updates account's phone number and roles (isMentor/isMentee), other account settings relating updates can be done through route **PATCH** `/account-settings/:account_id/`
- added tests for changing and checking account's password


mesh/accounts/views.py
```diff
@@ -152,109 +183,76 @@ def delete(self, request, account_id):
...
-                )
+        try:
+            data = json.loads(request.body)
+            email = data.get("email")
+            password = data.get("password")
+
+            # Query to check if user exists
+            account = Account.objects.get(email=email)
+            if account and account.encryptedPass == decrypt(password, account.salt):
+                return JsonResponse({"message": "Access granted"}, status=200)
```
misslame, line 196: This is very nice improvement to this function

mesh/tests/accounts_tests.py
```diff
@@ -106,12 +108,80 @@ def test_patch_update_account(self):
...
+        response = self.client.patch('/accounts/change-password/', json.dumps(change_password_data), content_type='application/json')
+        self.assertEqual(response.status_code, 204)
+
+        # verify if the password has indeed been changed b y attempting to log in with the new password
+        login_data = {
+            'email': 'test_change_password@email.com',
+            'password': 'new_secure_password'
+        }
+        login_response = self.client.post('/accounts/check-password/', json.dumps(login_data), content_type='application/json')
+        self.assertEqual(login_response.status_code, 200)
```
misslame, line 184: YES TESTING!

misslame (approved):
Love your exception handling and testing, I approve

mesh/tests/accounts_tests.py
```diff
@@ -31,17 +35,17 @@ def test_get_all_accounts(self):
         """
         Test case for getting all accounts. 
 
-        A GET request is sent to the '/api/accounts/' endpoint.
+        A GET request is sent to the '/api/accounts' endpoint.
         The test passes if the response status code is 200.
         """
-        response = self.client.get('/accounts/')
+        response = self.client.get('/accounts')
```
GuyWhoCode, line 41: On my end, this test only passes if I put `/accounts/`. Can you confirm if it passes on your end?

mesh/accounts/views.py
```diff
@@ -44,8 +57,19 @@ def post(self, request):
...
-            # data = json.loads(request.body)
-            salt,hash = encrypt(data['password'])
+
+            # check if there is any account created with this email/phoneNum
+            created_email = Account.objects.filter(email = data["email"]).values()
+            created_phone = Account.objects.filter(phoneNum = data["phoneNum"]).values()
+
+            if created_email or created_phone:
+                return JsonResponse(
+                    data={ "status":"account with same email or phone number exists" },
```
GuyWhoCode, line 67: For the sake of consistency, should the key of this JSONResponse be `error` instead of `status`?

GuyWhoCode (changes requested):
Looks pretty good overall but there's some minor tweaks that need to be done

mesh/tests/accounts_tests.py
bluesimp1102: yess you are right. For some reasons, it also accepts both `/accounts` when i run the test functions. But when I actually tried to send the request to the API, it only accepts the `/accounts/`. I changed the documentation to `/accounts/`

GuyWhoCode (approved):
LGTM

PR #339, opened 2024-05-21
Leaf/jack/login authentication

misslame:
- added property `last_login` to be able to use class Account with Django login and auth feature
- Set up cookies and Authentication Backend to handle login, logout, and session requests
- change new login route `/auth/login/` in the frontend


blu3eee (approved):

mesh/auth/views.py
```diff
@@ -0,0 +1,83 @@
...
+            return JsonResponse({"message": "Access granted", "is_logged_in": True, "account": account_data}, status=200)
+        else:
+            # Generic error message for security
+            return JsonResponse({"error": "Invalid credentials"}, status=401)
+    except json.JSONDecodeError:
+        return JsonResponse({"error": "Invalid JSON format"}, status=400)
+    except Exception as e:
+        return JsonResponse({"error": str(e)}, status=500)
+
+def logout_view(request):
```
shotoh, line 49: should there be a decorator requiring only POST requests?

mesh/auth/backend.py
```diff
@@ -0,0 +1,102 @@
...
+
+        Returns:
+            A dictionary containing the account data and its settings if they exist,
+            otherwise None.
+
+        If the account does not have settings, a default Settings instance is created.
+        """
+        from django.db import transaction
+        from mesh.accountSettings.models import Settings
+        with transaction.atomic():
```
shotoh: if you want I believe you can take this out and use the @transaction.atomic decorator, up to you

shotoh (approved):
looks good, I left a few comments but I'm just nitpicking at this point

PR #340, opened 2024-05-21
Leaf/ci django

misslame:
This will run django tests on every push related to backend (pip files, `manage.py` and `/mesh` folder) and every pull requests (targets can be changed)

Currently, the setup will run every test on the folder `/mesh/tests/` that ends with `_tests.py` 

- Modified `settings.py`: according to what I can look up online, `STATICFILES_DIRS` in `settings.py` is used to define additional directories where Django should look for static files (CSS, JavaScript, images, etc.) during the static file collection process. This might not be necessary as we're only using Django for backend API (correct me if I'm wrong), so uncomment this might not affect our current project setup.

### DIFF
--- a/.github/workflows/django-ci.yml
+++ b/.github/workflows/django-ci.yml
@@ -27,11 +27,11 @@ jobs:
         uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install Dependencies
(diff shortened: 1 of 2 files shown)


blu3eee (approved):

shotoh (approved):
looks good

Issue #336, opened 2024-05-05
[DevOps] Improve hot reloading for Django Channels with Docker


blu3eee:
Currently, when running the Django project with Docker and using Django Channels, we need to start the Django server with `daphne` (`daphne -b 0.0.0.0 -p 8000 mesh.asgi:application`) instead of the regular `python manage.py runserver`. This is because Django Channels requires running the ASGI server instead of the WSGI server.

`daphne` server does not support hot reloading on code changes, which makes the development workflow less efficient when working with Docker containers.

We need to explore better solutions to enable hot reloading for Django Channels when developing with Docker. Some potential solutions to consider is switching to `watchgog` or `uvicorn`

blu3eee:
resolved with pr #344

PR #344, opened 2024-05-24
Instant messaging with WebSocket

blu3eee:
**[Backend]**
- Apply changes in database models for conversations, messages, and conversation participants
- API endpoints for `/conversations` (create conversations, get conversation messages, and get users/accounts' conversations)
- Added tests for said endpoints

**[DevOps]**
- Changed to `uvicorn` from `daphne` to run the backend ASGI app for hot-reloading (see #336)


mesh/conversation/views.py
```diff
@@ -1,3 +1,170 @@
...
+        """
+        try:
+            REQUIRED_FIELDS = ['participants', 'conversationType']
+            data = validate_json_and_required_fields(request.body, REQUIRED_FIELDS)
+            
+            participants = data['participants']
+            conversationType = data['conversationType']
+
+            if not isinstance(participants, list):
+                return HttpResponseBadRequest("Invalid data")
```
shotoh: why not use a JsonResponse with an error message like the ones below?

mesh/conversation/views.py
```diff
@@ -1,3 +1,170 @@
...
+
+            # Create the conversation
+            conversation = Conversation.objects.create(conversationType=conversationType)
+
+            # Add participants to the conversation
+            try:
```
shotoh: this inner try/catch can be removed and the Account.DoesNotExist exception can be moved to the first try/catch

mesh/tests/chat_wsclient_pytests.py
```diff
@@ -0,0 +1,109 @@
...
+    conversationID = conversation.conversationID
+    
+    # Create test accounts asynchronously
+    account1 = await sync_to_async(Account.objects.create)(email='user1@example.com')
+    account2 = await sync_to_async(Account.objects.create)(email='user2@example.com')
+
+    # Add accounts to the conversation as participants
+    response = client.post(
+        '/conversations/', 
+        json.dumps({
```
shotoh: I believe the json.dumps method is redundant because of the content_type, not entirely sure

mesh/conversation/views.py
```diff
@@ -1,3 +1,170 @@
...
+                    "participants": { "accountID": int, "name": str }[]
+                }[]
+            }
+        - status=404: If the specified account is not found.
+        - status=500: If an internal server error occurs.
+    """
+    try:
+        accountID = account_id
+        # Get the account object
+        account = get_object_or_404(Account, accountID=accountID)
```
shotoh: I believe the Account.DoesNotExist exception doesn't occur at all with this? Or you could replace it with Account.objects.get and keep the exception

shotoh (commented):
left a few comments

Pipfile
```diff
@@ -16,9 +16,12 @@ django-cors-headers = "~=4.3.1"
 
 channels = "*"
 channels_redis = "*"
-daphne = "*"
```
misslame, line 19: what is daphne?

Pipfile
```diff
@@ -16,9 +16,12 @@ django-cors-headers = "~=4.3.1"
...
+uvicorn = "*"
```
misslame, line 18: Good choice

mesh/conversation/views.py
```diff
@@ -1,3 +1,170 @@
...
+                    "conversationID": int,
+                    "conversationType": str,
+                    "conversationName": str or None,
+                    "participants": { "accountID": int, "name": str }[]
+                }[]
+            }
+        - status=404: If the specified account is not found.
+        - status=500: If an internal server error occurs.
+    """
+    try:
```
misslame, line 110: I wonder how performant this is with large amounts of data. Not really an issue but maybe something to think about in order to have an engineering mindset.

misslame (approved):
Did you update dependencies? i have never seen this view in github before kinda cool. I will approve for now but shotoh had good questions about the exceptions.

mesh/conversation/views.py
blu3eee: Yes, I thought about it and added a todo comment as well as an issue for the matter (#349)

Pipfile
blu3eee: So `daphne` is an `ASGI` server used with Django Channels to handle WebSocket connections and other asynchronous features. Before this, we used `python manage.py server` to run the backend server, this command will run a `WSGI` server. By adding daphne, we allow the server to handle asynchronous communication and real-time data exchange with WebSocket

shotoh (approved):

PR #347, opened 2024-05-24
Leaf/A4.8 HTTP Post Call Handling

shotoh:
- added class view for tags
- updated GET endpoint
- added POST endpoint
- updated tests


blu3eee (approved):
LGTM!

misslame (approved):
Looks good, thank you for making a class. It is interesting very non-python like but I like it

PR #355, opened 2024-06-11
Use SVGR for svg components

blu3eee:
Adopt `svgr` to allow svgs to be loaded as React components for faster loading

**I also renamed some files that I missed from my last two PRs (just my own nitpicks in files naming).


blu3eee:
atm, I have a workaround by optimizing the svg files with the command using [`svgo`](https://github.com/svg/svgo). tdlr, `svgr` is also using `svgo` to optimize the svg files before "making" them React components. 

The problem with the config is that the svgr will allow us to import svg files to react components but not optimized (it is expected to optimized with `svgo` plugin by default). I'm blocked at this point with trying to make `svgr` to optimize the svg files on components importing instead of optimizing svg files by running the svgo cmd line ourselves. This workaround still works in terms of svg optimization but not the best way to do it. I tried to look around the open and closed issues in `svgr` but none of the solution works. I'm not too familiar with webpack configuration myself so I might have missed something.

blu3eee:
** This PR is ready to be merged to use optimized SVG instead. Further research is needed to make `svgr` optimize svgs automatically with `svgo` but that is another problem in the future.

SVGs like `MeshPfp.svg` and `MeshLogo.svg` when optimizing with `svgo` is not working as expected so I excluded those from the optimizing command.

Darren52BT (approved):
looks good to me

misslame (approved):

PR #356, opened 2024-06-12
Bug Fix: Adapt Profile Page Theming

### LINKED ISSUE #289: BUG: Adapt Profile Page To Use Theming
Adapt the Profile Page to use the MUI theme defined in the theme folder.

Darren52BT:
addresses #289.

This implements color theming for the profile page.

## What I did
-  In order to have the colors respond to the color mode, I have to reference them directly in the component files, so I commented out the colors in the profile page's css file and did just that.
- I also added a color for the profile picture edit UI in the MUI theme called buttonBackground, also used this for delete confirmation button.
- for the sweetalert being used for asking confirmation to delete a profile picture, since it wasn't possible to reference the MUI theme directly here I had to grab the current theme mode and list/use a set of hex code colors depending on the mode. 
- for some reason the text colors were missing for the light theme so I added them back.

## Concerns
- The profile page's sizing is not responsive. I think I was only supposed to address coloring here, so I can make up another issue for this later. This could be part of a larger story later since I think some of the other pages aren't responsive either.
- I will also make up an issue for adding back the cocogoose font class to everything that needs it on every page. I'll also look into general theming issues in general and decide if I should make that a larger issue.
## What you should see
Light Mode
image
image
image

Dark Mode
image
image
image


misslame:
responsive and reactive I think will be addressed with our a11y epic

meshapp/src/pages/Profiles/Page.tsx
```diff
@@ -20,22 +13,6 @@ import { axiosInstance } from "src/config/axios-config";
 // New Imports
```
misslame: i know you didn't add this comment, but this is a pointless comment, could you remove it  lol

meshapp/src/pages/Profiles/Page.tsx
```diff
@@ -261,20 +251,23 @@ const ProfileRole = (props: { isMentor: boolean; isMentee: boolean }) => {
...
-            marginBottom: "20px",
-            fontSize: "22px",
-          }}
-        >
-          Experience
-        </Typography>
-        <TestComponent />
-      </Box>
-    </ThemeProvider>
+    // <ThemeProvider theme={theme}>
```
misslame: why are there comments here? Will they come back?

misslame (commented):
I don't think I understand why themeprovider is being removed from all the steps or why it was there in the beginning. Shouldn't themeprovider be applied to the whole site/app?

meshapp/src/pages/Profiles/Page.tsx
Darren52BT: yeah for sure

meshapp/src/pages/Profiles/Page.tsx
Darren52BT: That's my bad I meant to remove that. And as for your question on why ThemeProviders were there and are being removed now, there is actually a global ThemeProvider and Theme being applied to the entire app currently; I think at the time Patrick was developing this this wasn't implemented yet so he made his own Theme and ThemeProviders to initially style them. Now that there is I removed them so I could properly adapt them to the global Theme.

misslame: ah okay yea a global theme is what I expect, thank you for the explanation, i was a bit confused.

misslame (approved):

blu3eee (approved):
lgtm!!

PR #359, opened 2024-06-14
API Endpoints refactoring

### LINKED ISSUE #358: [REFACTOR] RESTful API endpoints
**User Story:**

As a developer, I want to refactor and reorganize the backend part of the project to follow to RESTful standards, ensuring a consistent and logical structure for the API endpoints.

**Details:**

**Example of refactoring profiles endpoint:**

- **Current endpoints:**
  ```python
  urlpatterns = [
      # GET /profiles/profile-picture
      # POST /profiles/profile-picture
      # PATCH /profiles/profile-picture
      # DELETE /profiles/profile-picture
      path("profile-picture", profile_views.ProfilePicturesView.as_view(), name="profile-picture"),
      
      # GET /profiles/biography/:account_id
      # POST /profiles/biography/:account_id
      path("biography/<int:account_id>", handler, name='biography'),

      # GET /profiles/profile-picture/:account_id
      path("profile-picture/<int:account_id>", handler, name="profile-picture-details"),

      # GET /profiles/user-name/:account_id
      # POST /profiles/user-name/:account_id
      path("user-name/<int:account_id>", handler, name="user-name"),
      
      # GET /profiles/preferred-name/:account_id
      # POST /profiles/preferred-name/:account_id
      path("preferred-name/<int:account_id>", handler, name="preferred-name"),
      
      # GET /profiles/preferred-pronouns/:account_id
      # POST /profiles/preferred-pronouns/:account_id
      path("preferred-pronouns/<int:account_id>", handler, name="preferred-pronouns"),
  ]
  ```

- **Proposed Structure:**
  ```python
    urlpatterns = [
      # GET /profiles/:account-id/profile-picture
      # POST /profiles/:account-id/profile-picture
      # PATCH /profiles/:account-id/profile-picture
      # DELETE /profiles/:account-id/profile-picture
      path('<int:account_id>/profile-picture/', handler, name="profile-picture"),

      # GET /profiles/:account-id
      # GET /profiles/:account-id?fields=
      # PATCH /profiles/:account-id
      path('<int:account_id>/', handler, name='profile-details'),
  ]
  ```

**Technical details:**

- Refactor/group the related endpoints to follow the proposed structure.
- Ensure that the backend API accepts requests with or without trailing slashes for all endpoints.
- Ensure thorough testing of all updated endpoints to confirm they work as expected.

Why do this?
This aims to improve the clarity, maintainability, and consistency of the API, making it easier for developers to understand and work with.

blu3eee:
Addresses #358 

## What I Did
- Refactored redundant endpoints into one with search queries (`/profiles/:account_id?fields=` & `account-settings/:account_id?fields=`)
- Removed `re_path` routes, which is returning a template `index.html` on every request to the endpoint that doesn't have a handler, which is causing append slash behavior to not working (more details below)
- Adjusted test cases to fit with the refactoring

## Removal of `re_path(r'.*', )` in `mesh/urls.py`
- This removal is to allow the appending slash behavior to work properly
- About appending slash behavior: with the current Django setup, Django is setting `APPEND_SLASH = true` as a default behavior. This allows the Django app to "append" slash on requests that don't have a trailing slash by permanently redirecting the request to the correct one. For example, we have an API endpoint at `/example/helloworld/`, when requesting to `/example/helloworld`, the Django app is expected to permanently redirect the request to `/example/helloworld/`, acting as an "append slash" behavior. But said `re_path` in `mesh/urls.py` is causing this redirect not to take place when a request to `/example/helloworld` is sent to the app, because the re_path takes priority and returns the user a template HTML file as a response instead of redirecting to `/example/helloworld/`. 
- By removing said `re_path`, the "append" slash behavior is working properly, allowing users to send requests to an endpoint with or without a trailing slash.

### Testing with and without trailing slash
- As mentioned above, the appending slash behavior, under the hood, is actually permanently redirecting requests to the correct one. So when we testing a request without the trailing slash, it will return a response with a status code 301
image

- In test cases:
 
**❌ Incorrect way of sending test request (without trailing slash):**
```python
from django.test import TestCase, Client
class YourTestClass(TestCase):
    def setUp(self):
        self.client = Client()
    def your_test_case(self):
        response = self.client.get("/accounts")
        self.assertEqual(response.status_code, 200) # this will not pass ❌
```
This will causing failures as the `response.status_code` is 301 instead of 200, because the app is permanently redirect.

**✅ Correct way of sending test requests (without trailing slash):**
```python
from django.test import TestCase, Client
class YourTestClass(TestCase):
    def setUp(self):
        self.client = Client()
    def your_test_case(self):
        response = self.client.get("/accounts", follow=True)
        self.assertEqual(response.status_code, 200) # this will pass ✅
        response = self.client.get("/accounts/")
        self.assertEqual(response.status_code, 200) # this will pass ✅
```

### DIFF
--- a/mesh/accountSettings/views.py
+++ b/mesh/accountSettings/views.py
@@ -1,10 +1,16 @@
+# Django
 from django.http import JsonResponse, HttpResponse
 from django.views import View
 from django.core.serializers import serialize
 from django.core.exceptions import ValidationError, ObjectDoesNotExist
+
+# Libraries
 import json
 
+# Models
 from .models import Settings, Account
+
+# Utils
 from ..utils.validate_data import validate_json_and_required_fields
 
 def display_theme(request, account_id):
@@ -87,17 +93,30 @@ class SettingsView(View):
             )
 
 class SettingsDetailView(View):
+    valid_fields = ['isVerified', 'verificationToken', 'hasContentFilterEnabled', 'displayTheme', 'is2FAEnabled']    
+
     def get(self, request, account_id, *args, **kwargs):
         """
         Handles GET requests when the client fetches for a specific account settings
         """
         try: 
+            fields = request.GET.get('fields')
+            if fields:
+                fields = [field for field in fields.split(',') if field in self.valid_fields]
+            else:
+                fields = self.valid_fields
+            if len(fields) == 0:
+                return JsonResponse({'error': 'Invalid fields'}, status=400)
+            
             settings = Settings.objects.get(accountID= account_id)
-            settings_detail = serialize('json', [settings])
+            settings_detail = { field: getattr(settings, field, None) for field in fields }
+            # Add accountID to the response data
+            settings_detail['accountID'] = account_id
+        
             return JsonResponse(settings_detail, safe = False, status=200)
         
         except Settings.DoesNotExist:
-            return JsonResponse({'error': 'Settings for this account do not exist'}, status=404)
+            return JsonResponse({'error': 'Setting for account not found'}, status=404)
         
     def patch(self, request, account_id):
         """
@@ -105,16 +124,19 @@ class SettingsDetailView(View):
         """
         try:
             setting = Settings.objects.get(accountID=account_id)
-        except Settings.DoesNotExist:
-            return JsonResponse({'error': 'Settings do not exist'}, status=404)
+            data = json.loads(request.body)
 
-        data = json.loads(request.body)
-        # Here, update the setting's attributes based on the data received
-        setting.isVerified = data.get('isVerified', setting.isVerified)
-        setting.verificationToken = data.get('verificationToken', setting.verificationToken)
-        setting.hasContentFilterEnabled = data.get('hasContentFilterEnabled', setting.hasContentFilterEnabled)
-        setting.displayTheme = data.get('displayTheme', setting.displayTheme)
-        setting.is2FAEnabled = data.get('is2FAEnabled', setting.is2FAEnabled)
-        # After updating, save the setting
-        setting.save()
-        return HttpResponse(status=204)
+            for field, value in data.items():
+                if field in self.valid_fields:
+                    setattr(setting, field, value)
+            
+            # After updating, save the setting
+            setting.save()
+            return HttpResponse(status=204)
+        except Settings.DoesNotExist:
+            return JsonResponse({'error': 'Setting for account not found'}, status=404)
+        except json.JSONDecodeError:
+            return JsonResponse({"error": "Invalid JSON format."}, status=400)
+        except KeyError:
+            return JsonResponse({"error": "Invalid fields in the request."}, status=400)
+
--- a/mesh/exampleapi/urls.py
+++ b/mesh/exampleapi/urls.py
@@ -1,4 +1,4 @@
-from django.urls import path, include
+from django.urls import path
 from . import views as exampleapi_views
 
 urlpatterns = [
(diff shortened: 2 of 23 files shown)


mesh/exampleapi/urls.py
```diff
@@ -1,4 +1,4 @@
-from django.urls import path, include
+from django.urls import path
 from . import views as exampleapi_views
 
 urlpatterns = [
```
shotoh, line 4: shouldn't there be a url for the check_buckets view you added?

shotoh (approved):
looks really good

mesh/exampleapi/urls.py
blu3eee: Yes, but I decided not to include it as some people might not know what it actually does and accidentally delete every file in the bucket. As for people who know what it does and need it to check the files, they can add and use it themselves.

mesh/exampleapi/views.py
```diff
@@ -12,4 +10,36 @@ def index(request):
 def hello_world(request):
     if request.method == "GET":
         print(request)
-        return HttpResponse("Hello World!")
\ No newline at end of file
+        return HttpResponse("Hello World!")
+    
+# This if for testing purposes and cleanup saved images from the testing requests
+# warning: only use this if you know what you're doing with backblaze buckets
+def check_buckets(request):
```
misslame, line 17: Why is this in example?

misslame (commented):
Just one question

mesh/exampleapi/views.py
blu3eee: Oh, I was testing the buckets to check if it works and I didn't know where else to put it

JRSiwiecki: Is this used anywhere? And if it isn't why does it need to exist if there needs to be a warning about using it?

mesh/accountSettings/views.py
```diff
@@ -87,34 +93,50 @@ def post(self, request):
...
         """
         try: 
+            fields = request.GET.get('fields')
+            if fields:
+                fields = [field for field in fields.split(',') if field in self.valid_fields]
+            else:
+                fields = self.valid_fields
+            if len(fields) == 0:
+                return JsonResponse({'error': 'Invalid fields'}, status=400)
+
```
JRSiwiecki, line 110: Just curious... wasn't the `validate_json_and_required_fields` made for this or is this one a special case that requires us to not use the util function?

JRSiwiecki (commented):
Small questions, ping me on discord when you take a look at them 🍡

mesh/accountSettings/views.py
blu3eee: tdlr: `validate_json_and_required_fields` was not made for this. the two use cases are different.

- `validate_json_and_required_fields` is to validate the request body. It takes in a `json` object, check if the request body has the required fields for the action (i.e. creating account will need enough information) and returns an indicator saying if the request body is valid for the action.

- The fields checking in this particular case is getting the `fields` param from the request, then parse it into an array and double-check it with valid fields to see if the requested fields are matched with return object available fields. (Kindof mimicking `graphQL` behavior in a sense)
    + For example, the account settings have these fields: `twoFactorEnabled`, `theme`, and `notificationSetting`. The `fields` in the request body will indicate how much information the backend is returning, in some cases, we only need partial information. So, instead of returning the whole accountSetting object from the database, we only return the requested fields. If in the request body, `fields="theme,twoFactorEnabled"`, only those information is returned in the response.

JRSiwiecki: Okay that makes sense, is there no need to split this into its own function as well or does this seem like it'd be a one-off operation?

PR #364, opened 2024-07-17
implemented view only profile page

Darren52BT:
This PR implements the view only version of the profile page. 
## Changes:
- added a boolean called ``viewOnly`` to the Profile type that is passed into the profile's ``Page`` component, and this in turn is passed down to every component that needs to be adapted to view only.
- used this boolean to perform conditional rendering between view only/editable components
- added a boolean parameter``isSecond`` to ``ProfileAccordionCombobox`` that represents whether it is the second combobox in a accordion to implement the difference in colors shown here
image
- added another color to the theme palette called ``input.borderActive`` for the Biography component's border

## What You Should See
image
image


meshapp/src/models/profile.d.ts
```diff
@@ -1,6 +1,7 @@
 // TODO: Update order of fields to match order in the Profile schema
 // NOTE: Add/modify/remove fields as needed, but consider what needs to be refactored in ../profile-page.tsx
 export type Profile = {
+  viewOnly:boolean;
```
JRSiwiecki: Such a small nitpick but there could be a space here

meshapp/src/pages/Profiles/Page.tsx
```diff
@@ -28,6 +29,7 @@ import ProfileHeader from "./ProfileHeader";
  * @param {Education} props.education - an array containing objects that each contain a degree,school, and description
  * @param {Experience} props.experience - an array contains objects that each contain a occupation name, organization, and description
  * @param {number} props.accountID - ID of Profile account
+ * @param {boolean} props.viewOnly - whether profile is view only or not
```
JRSiwiecki: Is there a reason this is written twice?

meshapp/src/pages/Profiles/ProfileAccordion.tsx
```diff
@@ -68,6 +68,7 @@ export default function ProfileAccordion(props: {
   alwaysOpen?: boolean;
   editHandler?: Function;
   deleteHandler?: Function;
+  viewOnly?: boolean | undefined;
```
JRSiwiecki, line 71: Is there a reason this an some others are able to be `undefined`?

meshapp/src/pages/Profiles/tests/profile-examples.ts
```diff
@@ -2,6 +2,7 @@ import { Profile } from "src/models/profile";
 
 export const exampleProfile: Profile = {
   accountID: 1,
+  viewOnly:true,
```
JRSiwiecki: Small space here 👯

JRSiwiecki (commented):
Really small misc changes and questions that I have, this looks good though 💯 ping me when you get a chance to respond

meshapp/src/pages/Profiles/ProfileAccordion.tsx
Darren52BT: I did this since I thought it would be redundant and tedious to have to insert a value for viewOnly for other accordions where their purpose is to be edited; I can change this if you want me to.

meshapp/src/pages/Profiles/Page.tsx
Darren52BT: that's my bad I was going over all the components that had viewOnly and was copy pasting it in and ig I did it twice lol

meshapp/src/pages/Profiles/ProfileAccordion.tsx
JRSiwiecki: Oh okay that makes sense, nah I think it's fine, just was curious as to why this one had the option... but I think that as long as a bunch of other properties don't do the same thing it shouldn't matter...

JRSiwiecki (approved):
🥇

